PR #18, opened 2015-02-05
Add a backwards compatible change for enabling lower casing

fjy:


src/main/java/com/metamx/common/parsers/JSONParser.java
```diff
@@ -68,23 +68,33 @@ public Object apply(JsonNode node)
...
+    this.parserConfig = parserConfig;
+  }
+
   @Deprecated
   public JSONParser(Iterable<String> fieldNames)
   {
     this(new ObjectMapper(), fieldNames, null);
   }
 
+  @Deprecated
```
fjy: i think we shouldn't have fieldNames and excludes at this level

fjy: I actually think we should rework the constructor to be
Objectmapper, fieldNames, jsonparseconfig

I don't think excludes should be at this level, or if they are, be in the jsonparseconfig

src/main/java/com/metamx/common/parsers/JSONToLowerParser.java
```diff
@@ -0,0 +1,82 @@
+package com.metamx.common.parsers;
```
xvrl: license

src/main/java/com/metamx/common/parsers/JSONToLowerParser.java
```diff
@@ -0,0 +1,82 @@
...
+    try {
+      Map<String, Object> map = new LinkedHashMap<>();
+      JsonNode root = objectMapper.readTree(input);
+
+      Iterator<String> keysIter = (fieldNames == null ? root.fieldNames() : fieldNames.iterator());
+
+      while (keysIter.hasNext()) {
+        String key = keysIter.next().toLowerCase(); // only difference from JSONParser parse()
+
+        if (exclude.contains(key)) {
```
xvrl: should exclude be lowercased too?

PR #21, opened 2015-03-04
Add and maybe start handler instead of add.

b-slim:
Exception will raise if the service did already start

Basically, what is happening is that we have implemented a class that uses the ScheduledExecutorFactory to generate its own executors. This is done after initialization, though, so the Lifecycle has already been started. The method that is being used here to register things on the Lifecycle can only be used before it has been started. This PR is changing it to the method that can work even if the Lifecycle has already been started.

The big difference in the API is that it throws an exception, which must now be handled. Now, the only way that an exception is thrown is if the start() method throws it, the start() method does nothing here so it will not actually get thrown. Either way, we just propagate.

### DIFF
--- a/src/main/java/com/metamx/common/concurrent/ExecutorServices.java
+++ b/src/main/java/com/metamx/common/concurrent/ExecutorServices.java
@@ -16,6 +16,7 @@
 
 package com.metamx.common.concurrent;
 
+import com.google.common.base.Throwables;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.metamx.common.lifecycle.Lifecycle;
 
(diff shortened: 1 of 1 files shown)


drcrallen:
@b-slim : Can you please explain when this case arrises?

b-slim:
@drcrallen this case arises when i am creating a threadpool within a druid module after it has been started
```java
Caused by: com.metamx.common.ISE: Cannot add a handler after the Lifecycle has started, it doesn't work that way.
at com.metamx.common.lifecycle.Lifecycle.addHandler(Lifecycle.java:153)
at com.metamx.common.lifecycle.Lifecycle.addHandler(Lifecycle.java:139)
at com.metamx.common.concurrent.ExecutorServices.manageLifecycle(ExecutorServices.java:40)
at com.metamx.common.concurrent.ScheduledExecutors$5.create(ScheduledExecutors.java:196)
```   
the actual call is done like this
```java
return new DruidModule(
        configSupplier.get(),
    scheduledExecutorFactory.create(configSupplier.get().getThreadPoolSize(), "TH-pool-%d")
    );

cheddar:
@b-slim need a better description on your PR ;)

Basically, what is happening is that we have implemented a class that uses the ScheduledExecutorFactory to generate its own executors.  This is done after initialization, though, so the Lifecycle has already been started.  The method that is being used here to register things on the Lifecycle can only be used before it has been started.  This PR is changing it to the method that can work even if the Lifecycle has already been started.

The big difference in the API is that it throws an exception, which must now be handled.  Now, the only way that an exception is thrown is if the `start()` method throws it, the `start()` method does nothing here so it will not actually get thrown.  Either way, we just propagate.

PR #22, opened 2015-03-06
Add Compression and File utilities for Zip and GZ handling

drcrallen:
I moved these from druid-api and added / expanded the functionality.


drcrallen:
I think I want to add one or two more options for the utilities. give me a min

drcrallen:
Added more

drcrallen:
get --> build in the docs shouldn't be a change. Will fix

drcrallen:
I'm adding a few more cases, closing until they are in

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,370 @@
...
+   */
+  public static long gunzip(InputStream in, File outFile) throws IOException
+  {
+    try (GZIPInputStream gzipInputStream = new GZIPInputStream(fix7036144(in))) {
+      final ByteSink sink = Files.asByteSink(outFile);
+      return sink.writeFrom(gzipInputStream);
+    }
+  }
+
+  public static InputStream fix7036144(final InputStream in)
```
fjy: can we call this something else and reference the bug?

drcrallen:
It was asked which parts are new.

A very large part is new. Here's the prior [CompressionUtil](https://github.com/druid-io/druid-api/blob/7a63e601955f80826b6bfdb02631b8b641e8ee0b/src/main/java/io/druid/utils/CompressionUtils.java) which did not really have retries or data consistency checks.

FileUtils is new

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Bug is referenced below, but I'll put it in the method comments and change the name.

drcrallen: Fixed.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,349 @@
...
+      final File outDir,
+      final Predicate<Throwable> shouldRetry,
+      Boolean cacheLocally
+  ) throws IOException
+  {
+    if (!cacheLocally) {
+      return unzip(inputStreamSupplier.openStream(), outDir);
+    }
+    final File tmpFile = File.createTempFile("compressionUtilZipCache", ZIP_SUFFIX);
+    tmpFile.deleteOnExit();
```
himanshug: from the discussion in https://github.com/druid-io/druid/pull/1132 , tmpFile.deleteOnExit() can potentially cause memory leak and should be removed from here as well.

drcrallen: Yep fixed, thanks!

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  public static String getGzBaseName(String fname){
+    if(isGz(fname)){
+      return fname.substring(0, fname.length() - GZ_SUFFIX.length());
+    }
+    throw new IAE("[%s] is not a valid gz file name", fname);
+  }
+
+  public static final String GZ_SUFFIX = ".gz";
```
fjy: minor nitpick, but can we define constants at the top of the file?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,291 @@
...
+        return false;
+      }
+      if(!otherDigest.getAlgorithm().equals(digest.getAlgorithm())){
+        throw new IAE("Digest algorithm did not match. Expected [%s] got [%s]", digest.getAlgorithm(), otherDigest.getAlgorithm());
+      }
+      return Arrays.equals(digestBytes, otherDigest.digest());
+    }
+
+    // Only works for immutable children contents
+    public synchronized Long size()
```
fjy: can this be primitive long?

fjy: instead of calculating size in this function, can we calculate it as we go? we can also avoid locking on the implicit this

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,291 @@
...
+      final MessageDigest digest;
+      try {
+        digest = MessageDigest.getInstance(digestAlgorithm);
+      }
+      catch (NoSuchAlgorithmException e) {
+        throw new ISE(e, "Somehow I don't know how to do algorithm [%s]", digestAlgorithm);
+      }
+      try (final FileInputStream fileInputStream = new FileInputStream(path.toFile())) {
+        try (final DigestInputStream digestInputStream = new DigestInputStream(fileInputStream, digest)) {
+          final byte[] buffer = new byte[1 << 10];
```
fjy: where does the size come from? can we declare it as a constant if it is constant?

src/test/java/com/metamx/common/CompressionUtilsTest.java
```diff
@@ -0,0 +1,502 @@
...
+        Assert.assertArrayEquals(expected, Arrays.copyOfRange(found, expected.length * 0, expected.length * 1));
+        Assert.assertArrayEquals(expected, Arrays.copyOfRange(found, expected.length * 1, expected.length * 2));
+        Assert.assertArrayEquals(expected, Arrays.copyOfRange(found, expected.length * 2, expected.length * 3));
+      }
+    }
+  }
+
+  @Test
+  // http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7036144
+  public void test7036144workarround() throws IOException
```
fjy: can we give this a better name :P

src/test/java/com/metamx/common/CompressionUtilsTest.java
```diff
@@ -0,0 +1,502 @@
...
+        Assert.assertArrayEquals(expected, Arrays.copyOfRange(found, expected.length * 1, expected.length * 2));
+        Assert.assertArrayEquals(expected, Arrays.copyOfRange(found, expected.length * 2, expected.length * 3));
+      }
+    }
+  }
+
+
+  @Test
+  // http://bugs.java.com/bugdatabase/view_bug.do?bug_id=7036144
+  public void test7036144StreamWorkarround() throws IOException
```
fjy: same here

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,291 @@
...
+  };
+  public static final Predicate<Throwable> MD5ERROR = new Predicate<Throwable>()
+  {
+    @Override
+    public boolean apply(Throwable input)
+    {
+      return input instanceof FileUtils.MD5MismatchException;
+    }
+  };
+  public static final Predicate<Throwable> ALWAYS = new Predicate<Throwable>()
```
fjy: any reason why this is needed vs using what is in Predicates?

src/main/java/com/metamx/common/FileUtils.java
drcrallen: What do you mean "calculate it as we go"? 

Right now it calculates at method call time for a couple of reasons:

1. It is possible through the current logic to define children independent of actual directory structure, which negates the capacity for piggybacking File 
1. It is possible to have children co-resident in a directory with other files or directories not part of the file copy.
1. It is possible that children change the number of children they have over time

As such, I figured lazy calculation is probably best. But looking at it again, if the size is eagerly calculated, then any child result still has the same limitations. And I don't think anything is really gained. I'll give eager calculation a shot and see how it turns out.

drcrallen: Doing this simply will cause two major changes:
1. No longer thread safe
2. Tree must be built from the leaves up

I think these are acceptable limitations given the use case. I can add that to the description.

drcrallen: Fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Fixed

src/test/java/com/metamx/common/CompressionUtilsTest.java
drcrallen: Fixed

src/test/java/com/metamx/common/CompressionUtilsTest.java
drcrallen: Fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Nope fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipInputStream;
+import java.util.zip.ZipOutputStream;
+
+/**
```
xvrl: can we remove empty comments?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+    return FileUtils.FileCopyResult.buildResult(outFile.toPath(), "MD5");
+  }
+
+
+  public static FileUtils.FileCopyResult gzip(final File inFile, final File outDir) throws IOException
+  {
+    return gzip(inFile, outDir, FileUtils.MD5ERROR);
+  }
+
+  public static Boolean isZip(String fName)
```
xvrl: why Boolean, and not boolean?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+    }
+    return fName.endsWith(ZIP_SUFFIX); // Technically a file named `.zip` would be fine
+  }
+
+  public static Boolean isGz(String fName)
+  {
+    if (Strings.isNullOrEmpty(fName)) {
+      return false;
+    }
+    return fName.endsWith(GZ_SUFFIX) && fName.length() > GZ_SUFFIX.length();
```
xvrl: why is this logic different than for .zip?

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: It's mentioned in the isZip function:

`.zip` is a valid zip file, but `.gz` is not valid for the gzip decompression because it has no filename to decompress to. There is actually a test for a file named `.zip` as well.

src/main/java/com/metamx/common/CompressionUtils.java
xvrl: also, this probably doesn't need to be public?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+
+
+  public static FileUtils.FileCopyResult gzip(final File inFile, final File outDir) throws IOException
+  {
+    return gzip(inFile, outDir, FileUtils.MD5ERROR);
+  }
+
+  public static Boolean isZip(String fName)
+  {
+    if (Strings.isNullOrEmpty(fName)) {
```
xvrl: File.getName never returns null, so this is technically not needed, and != null is enough since the endWith will catch the empty case.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: I find boxed objects much more useful than their primitive types, and usually default to using them unless it's a regularly updated item like a loop index, or an array item. Will change anyways because all the other druid devs seem to hate boxed primitives :)

drcrallen: fixed

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+  }
+
+  public static long zip(File directory, OutputStream out) throws IOException
+  {
+    if (!directory.isDirectory()) {
+      throw new IOException(String.format("directory[%s] is not a directory", directory));
+    }
+    final File[] files = directory.listFiles();
+    if (files == null) {
+      throw new IAE("Directory not found [%s]", directory.getAbsolutePath());
```
xvrl: does directory.isDirectory() not catch this case already?

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Since this is a library function I tried to
1. Make the inputs work for crazy inputs, not just the normal use case
2. Make the logic as obvious as possible to anyone viewing the source.

I could also make it a big boolean expression on one line, but I don't like having logic that takes me more than 1 second to figure out exactly what its doing.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: After looking at isDirectory() docs, yes

drcrallen: I was relying too much on IntelliJ orange boxes for this one :)

drcrallen: fixed

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,348 @@
...
+    try (final ZipOutputStream zipOut = new ZipOutputStream(out)) {
+      for (File file : files) {
+        log.info("Adding file[%s] with size[%,d].  Total size so far[%,d]", file, file.length(), totalSize);
+        if (file.length() >= Integer.MAX_VALUE) {
+          zipOut.finish();
+          throw new IOException(String.format("file[%s] too large [%,d]", file, file.length()));
+        }
+        zipOut.putNextEntry(new ZipEntry(file.getName()));
+        try (InputStream inputStream = new FileInputStream(file)) {
+          totalSize += ByteStreams.copy(inputStream, zipOut);
```
xvrl: Looks like this used to be `ByteStreams.copy(Files.newInputStreamSupplier(file), zipOut)` and you changed it because newInputStreamSupplier is deprecated. Instead of adding the `try{}` block you can instead use `new ByteStreams.copy(new FileByteSource(file), zipOut)` to keep things simple.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: It is public because if a file name returns true from that function, then the library should be able to use it just fine. 

I had a lot of checks in various functions that had pretty much this exact same function. So instead of having it copied everywhere I have them using this method instead.

src/main/java/com/metamx/common/CompressionUtils.java
xvrl: as long as we provide an output file, the input should not matter, no?

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed via `totalSize += Files.asByteSource(file).copyTo(zipOut);`

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: That is true, but `CompressionUtils.gunzip(File, File)` takes an output *directory* to be consistent with `CompressionUtils.unzip(File, File)`

Any of the other gunzip methods (stream based) take a target file as the second parameter.

drcrallen: gunzip(File, File) could auto-detect if the target is a directory or not, and if it is not a directory treat it as a file to copy to.

xvrl: It seems a bit strange to try to maintain the same semantics for gzip and zip, since gzip is just a compression format and zip is a container format. tar.gz would be the closest to a zip file.

Are we actually using the gzip stuff anywhere yet?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+        totalSize += Files.asByteSource(file).copyTo(zipOut);
+      }
+      zipOut.closeEntry();
+    }
+
+    return totalSize;
+  }
+
+  public static FileUtils.FileCopyResult unzip(
+      final ByteSource inputStreamSupplier,
```
xvrl: maybe rename this to source, since InputStreamSuppliers are getting deprecated in guava?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+      final File outDir,
+      final Predicate<Throwable> shouldRetry,
+      Boolean cacheLocally
```
xvrl: We typically prefer primitives unless there is a need to have null values.
Using boxed primitives requires defining the behavior in case of null values.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+        log.warn("Could not delete zip cache at [%s]", tmpFile.getAbsolutePath());
+      }
+    }
+  }
+
+  public static FileUtils.FileCopyResult unzip(
+      final ByteSource inputStreamSupplier,
+      final File outDir,
+      Boolean cacheLocally
+  ) throws IOException
```
xvrl: the order of arguments for the version without retry predicate is inconsistent with the one with predicate, we should put the predicate last I think instead of adding it somewhere in the middle

drcrallen:
@xvrl Added a new commit which allows target outputs of gunzip to be files or directories

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: reasonable, will change

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+      zipOut.closeEntry();
+    }
+
+    return totalSize;
+  }
+
+  public static FileUtils.FileCopyResult unzip(
+      final ByteSource inputStreamSupplier,
+      final File outDir,
+      final Predicate<Throwable> shouldRetry,
```
xvrl: typically Throwables should not be caught, since you may not be able to recover from an Error, would Exception or IOException be enough in this case? Our method only throws IOException anyway.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: This simply bubbles up the logic of RetryUtil, which is also used in other tools I've seen.

I actually use this elsewhere if a thing like HDFS throws something that doesn't inherit IOException for example

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -35,6 +37,7 @@
   // The default buffer size to use (from IOUtils)
   private static final int DEFAULT_BUFFER_SIZE = 1024 * 4;
 
+  @Deprecated // Use FileUtils.retryCopy
```
xvrl: FileUtils.retryCopy only takes a ByteSource, it cannot work on InputStream directly, so we cannot deprecate it if we don't offer the same functionality.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -57,6 +60,7 @@ public static void copyToFileAndClose(InputStream is, File file, long timeout) t
     }
   }
 
+  @Deprecated // use FileUtils.retryCopy
```
xvrl: same here.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: To expand on this, The base logic (which this PR does not change) is to check the exception, and if it is deemed recoverable, then try again, and if not, throw it on up. It will wrap it as a RuntimeException if it is not one.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -81,4 +85,21 @@ public static void copyWithTimeout(InputStream is, OutputStream os, long timeout
       os.write(buffer, 0, n);
     }
   }
+
+  // Return a message digest of the input stream of copy from input stream to output stream. DOES NOT close inputStream
+  public static MessageDigest md5StreamToFile(InputStream inputStream, File outFile) throws IOException
```
xvrl: to keep in line with existing method names, maybe copyToFileWithDigest might be more consistent?
Maybe allowing to pass in the message digest would make this more flexible and not just limited to md5?

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: The goal was to try and stop using InputStream and instead use ByteSource so that retries can even be attempted. Using InputStream in a way such that the operation cannot be attempted again is something IMHO we should try to stop in general. That is why I deprecated this.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+import java.util.zip.GZIPOutputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipInputStream;
+import java.util.zip.ZipOutputStream;
+
+public class CompressionUtils
+{
+  private static final Logger log = new Logger(CompressionUtils.class);
+  private static final int DEFAULT_RETRY_COUNT = 10;
```
xvrl: 10 seems a little excessive for number of default retries, no?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+            FileUtils.retryCopy(
+                new ByteSource()
+                {
+                  @Override
+                  public InputStream openStream() throws IOException
+                  {
+                    return new BufferedInputStream(zipFile.getInputStream(entry));
+                  }
+                },
+                FileUtils.MD5ERROR,
```
xvrl: I'm not sure what we are trying to guard against with retries here. When trying to retry a read from file, if the checksums don't match the first time, it's unlikely they will match the second time, no? Otherwise, this would indicate memory corruption issues, in which case we have much bigger problems.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,326 @@
...
+    if (Strings.isNullOrEmpty(fName)) {
+      return false;
+    }
+    return fName.endsWith(GZ_SUFFIX) && fName.length() > GZ_SUFFIX.length();
+  }
+
+  public static String getGzBaseName(String fname)
+  {
+    if (isGz(fname)) {
+      return fname.substring(0, fname.length() - GZ_SUFFIX.length());
```
xvrl: we can probably just use Files.getNameWithoutExtension here?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,302 @@
...
+
+  /**
+   * Retry copy attempts from input stream to output stream. Does *not* check to make sure data was intact during the transfer
+   *
+   * @param inputStreamSupplier  Supplier for input streams to copy from. The stream is closed on every retry.
+   * @param shouldRetry          Predicate to determine if the throwable is recoverable for a retry
+   * @param outputStreamSupplier Supplier for output streams. The stream is closed on every retry.
+   * @param maxAttempts          Maximum number of retries before failing
+   */
+  public static long retryCopy(
```
xvrl: having two methods with the same name but quite different signatures seems a bit confusing.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Other retries had default of 10 so I just copied that behavior here. I personally think 3 is more reasonable.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: I was considering moving the ByteSource/ByteSink one to StreamUtils. Would that help?

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: It is making sure from the input stream to the resultant file is consistent. A mismatch of MD5 at this stage would probably be due to problems in getting an entry persisted to disk properly. If a file->file copy is not something that needs any checks at all then that's fine, I'll just make sure its commented in the method description that it assumes roses with regards to zip extraction.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: changed to 3

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: fixed

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Yes, in QTL

drcrallen:
Commits squashed.

@xvrl  Are you still reviewing this or have I addressed all your concerns?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,302 @@
...
+
+  /**
+   * Retry copy attempts from input stream to output stream. Does *not* check to make sure data was intact during the transfer
+   *
+   * @param byteSource  Supplier for input streams to copy from. The stream is closed on every retry.
+   * @param byteSink    Supplier for output streams. The stream is closed on every retry.
+   * @param shouldRetry Predicate to determine if the throwable is recoverable for a retry
+   * @param maxAttempts Maximum number of retries before failing
+   */
+  public static long retryCopy(
```
xvrl: I feel it's confusing to have to methods with the same name, with very similar signatures to have very different behavior. The behavior should be more or less clear from the name. It's also confusing why one would take a ByteSink and the other one a File, one return a long, the other one a FileCopyResult. I feel this should be two different methods.

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,302 @@
...
+          shouldRetry,
+          maxAttempts
+      );
+    }
+    catch (Exception e) {
+      throw Throwables.propagate(e);
+    }
+  }
+
+  public static class MD5MismatchException extends IOException
```
xvrl: maybe something more generic like ChecksumException (like hadoop) would be more useful?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    public static FileCopyResult buildResult(Path path, FileCopyResult... children) throws IOException
+    {
+      return buildResult(path, "MD5", children);
+    }
+
+    public static FileCopyResult buildResult(
+        Path path,
+        String digestAlgorithm,
```
xvrl: taking MessageDigest objects instead of just String might make the API more explicit and less confusing as to what algorithms are supported. It would also remove all the unnecessary checks around getting an instance of the digest?

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -44,9 +48,12 @@ public static void copyToFileAndClose(InputStream is, File file) throws IOExcept
     finally {
       CloseQuietly.close(is);
     }
+    return FileUtils.FileCopyResult.buildResult(file.toPath());
```
xvrl: I am concerned about the addition of buildResult here, since this will cause every single copy to read the data twice for the copy and for the checksum, instead of once as before, which could be expensive when dealing with large files and cause performance regressions.

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,302 @@
...
+    }
+  }
+
+  // Keeps results of a file copy, including children and total size of this result and any children result.
+  // This class is NOT thread safe.
+  // Child size is eagerly calculated and any modifications to the file after the child is added are not accounted for.
+  // As such, this result and all child results should be considered immutable and should be built from the leafs up.
+  public static class FileCopyResult
+  {
+    private static final int DEFAULT_BUFFER_SIZE = 1 << 10;
```
xvrl: re-use DEFAULT_BUFFER_SIZE from StreamUtils?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,302 @@
...
+   * Checks the given file against the passed message digest
+   *
+   * @param file        The file to check
+   * @param otherDigest The expected digest results. otherDigest.digest() is called
+   *
+   * @return `true` if the digests match, `false` if they do not
+   *
+   * @throws IOException passed from inner stream errors
+   */
+  public static Boolean checkDigest(final File file, final MessageDigest otherDigest) throws IOException
```
xvrl: if we never return null, this can be just a simple boolean

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: Reverted, but still made it return `long` since most copy functions anywhere return at least the size of what was copied.

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Fixed, also made DEFAULT_BUFFER_SIZE protected

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Moved one to StreamUtils

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,372 @@
...
+   * @throws IOException
+   */
+  public static FileUtils.FileCopyResult gunzip(final File pulledFile, File out) throws IOException
+  {
+    if (out.isDirectory()) {
+      final String name = pulledFile.getName();
+      if (!isGz(name)) {
+        throw new IAE("[%s] is not a .gz file", pulledFile.getAbsolutePath());
+      }
+      final File outFile = Paths.get(out.getAbsolutePath(), getGzBaseName(name))
```
xvrl: why not just use `new File(out, basename)`? getAbsolutePath() is technically not always available

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,372 @@
...
+                                .toFile();
+      log.info("Gunzipping file[%s] to [%s]", pulledFile.getAbsolutePath(), outFile.getAbsolutePath());
```
xvrl: no need for absolutePath

src/main/java/com/metamx/common/FileUtils.java
drcrallen: I put in a commit to change / expand how this is handled. I'm not convinced it is cleaner but am willing to go with it if you like it better.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Because I had forgotten at the time what the magic File syntax was. Changed.

But related to your getAbsolutePath comment: There are a lot of places in debugging statements that use getAbsolutePath. Is a simple toString() the preferred way to go when printing debug info on File objects?

xvrl:
In general, I feel we try to get out of our way to provide flexible checksums in FileCopyResult, but at the end of the day, methods like unzip(InputStream, File) only return md5, which seems a bit limiting.

Either we should provide good support for checksumming throughout, or not support it at all as part of the compression utilities and provide separate utility methods to do checksumming separately. Both gzip and zip already do checksumming internally (gzip we just need check the length for concatenated files, so it might not be as useful to support it there.

src/main/java/com/metamx/common/CompressionUtils.java
xvrl: most command like tools typically report errors in terms of relative paths, not absolute paths, unless necessary. Absolute paths can become terribly confusing because you have to decide whether to resolve symbolic links or not, which is something better left to the user.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: fixed

drcrallen:
After a long conversation with @xvrl  I am in agreement that the checksum approach previously presented did not solve enough of a problem, and any solution implemented at the level I was going for probably isn't going to be flexible enough in a reasonable manner. As such checksumming is left out of this implementation to be exercised by the callers if needed, to be verified against their source of truth as they desire, and to be implemented on the portion of the data which is deemed at risk.

To have all those considerations in the code and not suffer problems with implementations that do not share the same set of risk criteria simply wasn't working out.

Checksumming has been removed from the PR.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,388 @@
...
+import java.util.Enumeration;
+import java.util.concurrent.Callable;
+import java.util.zip.GZIPInputStream;
+import java.util.zip.GZIPOutputStream;
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipFile;
+import java.util.zip.ZipInputStream;
+import java.util.zip.ZipOutputStream;
+
+public class CompressionUtils
```
gianm: I think all the functions in this class need at least a little bit of documentation, especially around what their return values mean, when they will or won't throw exceptions, and what the retry behavior is.

Also, do we actually need all of these? More functions are more convenient but also more code to test and maintain.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,388 @@
...
+            },
+            shouldRetry,
+            DEFAULT_RETRY_COUNT
+        );
+      }
+      catch (Exception e) {
+        throw Throwables.propagate(e);
+      }
+    }
+    final File tmpFile = File.createTempFile("compressionUtilZipCache", ZIP_SUFFIX);
```
gianm: Putting this in an else block would be clearer.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,388 @@
...
+    }
+  }
+
+  public static FileUtils.FileCopyResult unzip(
+      final ByteSource byteSource,
+      final File outDir,
+      boolean cacheLocally
+  ) throws IOException
+  {
+    return unzip(byteSource, outDir, Predicates.<Throwable>alwaysTrue(), cacheLocally);
```
gianm: I think that by default we should retry on Exceptions but not on other kinds of Throwables.

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,388 @@
...
+                Predicates.<Throwable>alwaysTrue(),
+                DEFAULT_RETRY_COUNT
+            ).getFiles()
+        );
+      }
+    }
+    return result;
+  }
+
+  // Use unzip(ByteStream, File) if possible
```
gianm: Does that mean this method should be deprecated?

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,388 @@
...
+   * @param pulledFile The source of the gz data
+   * @param out        Either a directory or a target file
+   *
+   * @return The result of the file copy
+   *
+   * @throws IOException
+   */
+  public static FileUtils.FileCopyResult gunzip(final File pulledFile, File out) throws IOException
+  {
+    if (out.isDirectory()) {
```
gianm: Is there a reason that we need to switch behavior based on whether or not "out" is a directory? I think those kinds of magical APIs make code more confusing, because it's not always clear from the call site whether the second parameter is supposed to be a directory or file, and that makes it harder to understand.

src/main/java/com/metamx/common/CompressionUtils.java
gianm: I think it's fine to ask users to always provide a filename when gunzipping to the filesystem. I agree with Xavier that it's not really worth trying keep the API identical between zip and gzip since they aren't the same kind of thing.

gianm: We might even want to stress that difference by calling the methods different things. i.e. instead of "gunzip" and "unzip", "gunzipToFile" and "unzipToDirectory". When in doubt, add more words :P

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,116 @@
...
+    {
+      this(files == null ? ImmutableList.<File>of() : Arrays.asList(files));
+    }
+
+    public FileCopyResult(Collection<File> files)
+    {
+      this.addSizedFiles(files);
+    }
+
+    private void addSizedFiles(Collection<File> files)
```
gianm: Why is this private, when addFiles does the same thing and is public?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -0,0 +1,116 @@
...
+    catch (Exception e) {
+      throw Throwables.propagate(e);
+    }
+  }
+
+  // Keeps results of a file copy, including children and total size of this result and any children result.
+  // This class is NOT thread safe.
+  // Child size is eagerly calculated and any modifications to the file after the child is added are not accounted for.
+  // As such, this result and all child results should be considered immutable and should be built from the leafs up.
+  public static class FileCopyResult
```
gianm: Those comments should be a javadoc.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -33,34 +40,38 @@
 public class StreamUtils
 {
   // The default buffer size to use (from IOUtils)
-  private static final int DEFAULT_BUFFER_SIZE = 1024 * 4;
+  protected static final int DEFAULT_BUFFER_SIZE = 1024 * 4;
```
gianm: Why is this protected?

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: No because there are many cases where a ByteStream would not be available. @xvrl and I had a discussion about that previously

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Because the intention is to let potential future enhancements extend the functionality of FileCopyResult (example: ChecksumedFileCopyResult) and might need to override `addFiles(...)`

I will make this `protected` for such a case though.

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: It was used in FileUtils and CompressionUtils in older versions of this PR. Will revert.

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Fixing

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: changed it to just file so as to not magically guess what the user wants.

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: A good example would be something like the inputstream of a POST

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixing

drcrallen: fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/CompressionUtils.java
drcrallen: Fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: fixed

src/main/java/com/metamx/common/FileUtils.java
drcrallen: fixed

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -0,0 +1,481 @@
...
+                DEFAULT_RETRY_COUNT
+            ).getFiles()
+        );
+      }
+    }
+    return result;
+  }
+
+  // Use unzip(ByteStream, File) if possible
+  public static FileUtils.FileCopyResult unzip(InputStream in, File outDir) throws IOException
```
xvrl: this one's missing javadocs for some reason

drcrallen: For some of the "older" functions I did not add java docs, but rather copied them directly from druid-api. I can add for these though

drcrallen: Fixed

xvrl: I think it makes sense to document those as well, since they are new to java-util. Otherwise, we would just leave them in druid-api. We also made some changes to the return values.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -35,32 +39,36 @@
   // The default buffer size to use (from IOUtils)
   private static final int DEFAULT_BUFFER_SIZE = 1024 * 4;
 
-  public static void copyToFileAndClose(InputStream is, File file) throws IOException
+  // It is highly advised to use FileUtils.retryCopy whenever possible, and not use a raw `InputStream`
+  // This should only be used if there is absolutely no way to replay the InputStream
```
xvrl: can we make those javadocs as well?

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -35,32 +39,36 @@
...
     }
     finally {
       CloseQuietly.close(is);
     }
   }
 
-  public static void copyToFileAndClose(InputStream is, File file, long timeout) throws IOException, TimeoutException
+  // It is highly advised to use FileUtils.retryCopy whenever possible, and not use a raw `InputStream`
+  // This should only be used if there is absolutely no way to replay the InputStream
+  public static long copyToFileAndClose(InputStream is, File file, long timeout) throws IOException, TimeoutException
```
xvrl: same here

PR #24, opened 2015-06-02
Add flush calls to StreamUtils to handle errors on buffered stream flushes

drcrallen:
The new unit tests fail without the extra flush calls.

### DIFF
--- a/src/main/java/com/metamx/common/StreamUtils.java
+++ b/src/main/java/com/metamx/common/StreamUtils.java
@@ -55,7 +55,9 @@ public class StreamUtils
   {
     file.getParentFile().mkdirs();
     try (OutputStream os = new BufferedOutputStream(new FileOutputStream(file))) {
-      return ByteStreams.copy(is, os);
+      final long result =  ByteStreams.copy(is, os);
+      os.flush();
+      return result;
     }
     finally {
       CloseQuietly.close(is);
@@ -79,7 +81,9 @@ public class StreamUtils
   {
     file.getParentFile().mkdirs();
     try (OutputStream os = new BufferedOutputStream(new FileOutputStream(file))) {
-      return copyWithTimeout(is, os, timeout);
+      final long retval = copyWithTimeout(is, os, timeout);
+      os.flush();
+      return retval;
     }
     finally {
       CloseQuietly.close(is);
@@ -99,7 +103,9 @@ public class StreamUtils
   public static long copyAndClose(InputStream is, OutputStream os) throws IOException
   {
     try {
-      return ByteStreams.copy(is, os);
+      final long retval = ByteStreams.copy(is, os);
+      os.flush();
+      return retval;
     }
     finally {
       CloseQuietly.close(is);
@@ -133,6 +139,7 @@ public class StreamUtils
       os.write(buffer, 0, n);
       size += n;
     }
+    os.flush();
     return size;
   }
 
@@ -163,7 +170,9 @@ public class StreamUtils
               try {
                 inputStream = byteSource.openStream();
                 outputStream = byteSink.openStream();
-                return ByteStreams.copy(inputStream, outputStream);
+                final long retval =  ByteStreams.copy(inputStream, outputStream);
+                outputStream.flush();
+                return retval;
               }
               finally {
                 CloseQuietly.close(inputStream);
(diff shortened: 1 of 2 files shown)


nishantmonu51:
do we need similar fixes in FileSmoosher also ?

nishantmonu51:
+1

drcrallen:
@nishantmonu51 : probably. I honestly only touched some methods I had put in PRs for previously, and didn't look everywhere.

drcrallen:
@nishantmonu51 : You should be able to merge this if you're ok with it.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -55,7 +55,9 @@ public static long copyToFileAndClose(InputStream is, File file) throws IOExcept
   {
     file.getParentFile().mkdirs();
     try (OutputStream os = new BufferedOutputStream(new FileOutputStream(file))) {
-      return ByteStreams.copy(is, os);
+      final long result =  ByteStreams.copy(is, os);
+      os.flush();
```
xvrl: the try with resources statement should flush and throw exceptions when closing the output stream. It shouldn't be necessary to flush here.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -99,7 +103,9 @@ public static long copyToFileAndClose(InputStream is, File file, long timeout) t
   public static long copyAndClose(InputStream is, OutputStream os) throws IOException
   {
     try {
-      return ByteStreams.copy(is, os);
+      final long retval = ByteStreams.copy(is, os);
+      os.flush();
```
xvrl: the reason the exception is getting swallowed is because we use CloseQuietly.close(os). Instead of flushing we should simply not swallow exceptions when closing the output or input stream.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -133,6 +139,7 @@ public static long copyWithTimeout(InputStream is, OutputStream os, long timeout
       os.write(buffer, 0, n);
       size += n;
     }
+    os.flush();
```
xvrl: I believe the flush should be left to the caller and not be part of this method. Whether the caller uses a buffered or unbuffered input stream should not be the concern of this method.

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -163,7 +170,9 @@ public Long call() throws Exception
               try {
                 inputStream = byteSource.openStream();
                 outputStream = byteSink.openStream();
-                return ByteStreams.copy(inputStream, outputStream);
+                final long retval =  ByteStreams.copy(inputStream, outputStream);
+                outputStream.flush();
```
xvrl: same here, remove CloseQuietly instead of calling flush

gianm:
I agree with @xvrl, if we want exceptions to be thrown on problems then we should close loudly, not quietly. IMO it makes sense to leave the decision of whether or not to force a flush to the caller.

src/main/java/com/metamx/common/StreamUtils.java
xvrl: we should add a comment that this is a workaround for JDK7 bug, that's been fixed in JDK8 http://hg.openjdk.java.net/jdk8/jdk8/jdk/rev/759aa847dcaf

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -79,7 +81,9 @@ public static long copyToFileAndClose(InputStream is, File file, long timeout) t
   {
     file.getParentFile().mkdirs();
     try (OutputStream os = new BufferedOutputStream(new FileOutputStream(file))) {
-      return copyWithTimeout(is, os, timeout);
+      final long retval = copyWithTimeout(is, os, timeout);
+      os.flush();
```
xvrl: add comment here as well

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: Won't work for filtered output stream due to http://hg.openjdk.java.net/jdk8/jdk8/jdk/rev/759aa847dcaf

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: Reasonable, removing in another PR

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: Adding comment about http://hg.openjdk.java.net/jdk8/jdk8/jdk/rev/759aa847dcaf

PR #25, opened 2015-06-02
Add flush calls in CompressionUtils to help ensure that flush errors are thrown properly

drcrallen:
This should not be merged until
https://github.com/metamx/java-util/pull/24
is in because the unit tests will not pass.

### DIFF
--- a/src/main/java/com/metamx/common/CompressionUtils.java
+++ b/src/main/java/com/metamx/common/CompressionUtils.java
@@ -99,6 +99,7 @@ public class CompressionUtils
         totalSize += Files.asByteSource(file).copyTo(zipOut);
       }
       zipOut.closeEntry();
+      zipOut.flush();
     }
 
     return totalSize;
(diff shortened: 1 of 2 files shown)


nishantmonu51:
+1

src/main/java/com/metamx/common/CompressionUtils.java
```diff
@@ -99,6 +99,7 @@ public static long zip(File directory, OutputStream out) throws IOException
         totalSize += Files.asByteSource(file).copyTo(zipOut);
       }
       zipOut.closeEntry();
+      zipOut.flush();
```
xvrl: I think it should be left to the caller to handle exceptions thrown when closing the outputstream and not rely on this method to flush.

drcrallen: output stream is closed via try with resources.

PR #27, opened 2015-08-11
avoid string conversion while doing timestamp parsing whenever possible

nishantmonu51:
avoid string conversion while doing timestamp parsing whenever possible.

### DIFF
--- a/src/main/java/com/metamx/common/parsers/TimestampParser.java
+++ b/src/main/java/com/metamx/common/parsers/TimestampParser.java
@@ -60,46 +60,29 @@ public class TimestampParser
           return new DateTime(ParserUtils.stripQuotes(input));
         }
       };
-    } else if (format.equalsIgnoreCase("posix")) {
+    } else if (format.equalsIgnoreCase("posix")
+               || format.equalsIgnoreCase("millis")
+               || format.equalsIgnoreCase("nano")) {
+      final Function<Number, DateTime> numericFun = createNumericTimestampParser(format);
       return new Function<String, DateTime>()
       {
         @Override
         public DateTime apply(String input)
         {
           Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          return new DateTime(Long.parseLong(ParserUtils.stripQuotes(input)) * 1000);
+          return numericFun.apply(Long.parseLong(ParserUtils.stripQuotes(input)));
         }
       };
     } else if (format.equalsIgnoreCase("ruby")) {
+      // Numeric parser ignores millis for ruby.
+      final Function<Number, DateTime> numericFun = createNumericTimestampParser(format);
       return new Function<String, DateTime>()
       {
         @Override
         public DateTime apply(String input)
         {
           Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          Double ts = Double.parseDouble(ParserUtils.stripQuotes(input));
-          Long jts = ts.longValue() * 1000; // ignoring milli secs
-          return new DateTime(jts);
-        }
-      };
-    } else if (format.equalsIgnoreCase("millis")) {
-      return new Function<String, DateTime>()
-      {
-        @Override
-        public DateTime apply(String input)
-        {
-          Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          return new DateTime(Long.parseLong(ParserUtils.stripQuotes(input)));
-        }
-      };
-    } else if (format.equalsIgnoreCase("nano")) {
-      return new Function<String, DateTime>() {
-        @Override
-        public DateTime apply(String input) {
-          Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          long timeNs = Long.parseLong(ParserUtils.stripQuotes(input));
-          // Convert to milliseconds, effectively: ms = floor(time in ns / 1000000)
-          return new DateTime(timeNs / 1000000L);
+          return numericFun.apply(Double.parseDouble(ParserUtils.stripQuotes(input)));
         }
       };
     } else {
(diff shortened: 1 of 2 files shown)


xvrl:
it might make sense to create a separate timestamp parser for Long values instead of lumping it together in the existing one. That'll preserve backwards compatibility and make it clearer I hope.

nishantmonu51:
have made changes to the api to preserve backwards compatible. 
Please have a look again.

src/main/java/com/metamx/common/parsers/TimestampParser.java
```diff
@@ -28,76 +28,98 @@
...
         @Override
-        public DateTime apply(String input)
+        public DateTime apply(Object input)
         {
-          Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          return new DateTime(Long.parseLong(ParserUtils.stripQuotes(input)) * 1000);
+          Preconditions.checkArgument(input != null, "null timestamp");
+          String str = input.toString();
+          Preconditions.checkArgument(!str.isEmpty(), "empty timestamp");
+          return new DateTime(Long.parseLong(ParserUtils.stripQuotes(str)) * 1000);
```
himanshug: why are we not doing `instanceof Number` check in this case?

src/main/java/com/metamx/common/parsers/TimestampParser.java
```diff
@@ -28,76 +28,98 @@
...
         @Override
-        public DateTime apply(String input)
+        public DateTime apply(Object input)
         {
-          Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          Double ts = Double.parseDouble(ParserUtils.stripQuotes(input));
+          Preconditions.checkArgument(input != null , "null timestamp");
+          String str = input.toString();
+          Preconditions.checkArgument(!str.isEmpty(), "empty timestamp");
+          Double ts = Double.parseDouble(ParserUtils.stripQuotes(str));
```
himanshug: why are we not doing instanceof Number check in this case?

src/main/java/com/metamx/common/parsers/TimestampParser.java
```diff
@@ -28,76 +28,107 @@
 {
   private static final Logger log = new Logger(TimestampParser.class);
 
-  public static Function<String, DateTime> createTimestampParser(
+  public static Function<Object, DateTime> createTimestampParserGeneric(
```
xvrl: I wonder if we should leave the existing one as is, and create new one that is specific to Long values. That would probably cleaner, remove a lot of branching in this method, and also leave the signature more specific as to which types we expect.

nishantmonu51: i tried doing that but its not same for all the cases, nano and ruby have different logic since the Number cannot be directly parsed to timestamp.

nishantmonu51:
@xvrl @himanshug have handled the review comments, please have a look.

src/main/java/com/metamx/common/parsers/TimestampParser.java
```diff
@@ -93,9 +93,11 @@ public DateTime apply(String input)
         }
       };
     } else if (format.equalsIgnoreCase("nano")) {
-      return new Function<String, DateTime>() {
+      return new Function<String, DateTime>()
```
himanshug: it appears that 4 "else if" for "posix", "ruby", "nano" and "mills" could be replaced by single else if like
```
..
else if (format.equalsIgnoreCase("posix"))  ||
           format.equalsIgnoreCase("ruby"))  ||
           format.equalsIgnoreCase("millis"))  ||
           format.equalsIgnoreCase("nano")) ) {
    return new Function<String, DateTime>() {
        private Function<Number, DateTime> func = createNumericTimestampParser(format);
        public DateTime apply(String input) {
            return func.apply(Long.parseLong(ParserUtils.stripQuotes(input)));
        }
    }
} 
else ...
```

nishantmonu51: done. refactored.

himanshug:
LGTM

src/main/java/com/metamx/common/parsers/TimestampParser.java
```diff
@@ -60,46 +60,28 @@ public DateTime apply(String input)
...
       {
         @Override
         public DateTime apply(String input)
         {
           Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
-          return new DateTime(Long.parseLong(ParserUtils.stripQuotes(input)) * 1000);
+          return numericFun.apply(Long.parseLong(ParserUtils.stripQuotes(input)));
         }
       };
     } else if (format.equalsIgnoreCase("ruby")) {
```
xvrl: any reason ruby is still separate from the other numeric ones?

drcrallen: Does that change fit within the scope of this PR, or would that be better for a different PR?

nishantmonu51: ruby supports Double but other formats only long.

xvrl: ok, I see it now. I'm surprised it's actually working given the fact that createNumericTimestampParser uses input.longValue() for ruby format as far as I can see.

nishantmonu51: the ruby parser igores millis. 
prev impl - 
 Preconditions.checkArgument(input != null && !input.isEmpty(), "null timestamp");
          Double ts = Double.parseDouble(ParserUtils.stripQuotes(input));
          Long jts = ts.longValue() * 1000; // ignoring milli secs // done in numeric parser
          return new DateTime(jts);     /// in numeric parser now. 

Refactoring it to be more clear now.

nishantmonu51: TimestampParserTest.testRuby verifies ruby implementation. 
Added comments in code.

PR #28, opened 2015-08-18
Make Lifecycle ignore repeated stop requests

drcrallen:
Related to https://github.com/druid-io/druid/pull/1627

### DIFF
--- a/src/main/java/com/metamx/common/lifecycle/Lifecycle.java
+++ b/src/main/java/com/metamx/common/lifecycle/Lifecycle.java
@@ -51,6 +51,7 @@ public class Lifecycle
 
   private final Map<Stage, CopyOnWriteArrayList<Handler>> handlers;
   private final AtomicBoolean started = new AtomicBoolean(false);
+  private final AtomicBoolean shutdownHookRegistered = new AtomicBoolean(false);
   private volatile Stage currStage = null;
 
   public static enum Stage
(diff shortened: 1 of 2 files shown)


src/main/java/com/metamx/common/lifecycle/Lifecycle.java
```diff
@@ -243,7 +250,10 @@ public void addMaybeStartHandler(Handler handler, Stage stage) throws Exception
   public void start() throws Exception
   {
     synchronized (handlers) {
-      started.set(true);
+      if (!started.compareAndSet(false, true)) {
```
gianm: Should this be an error? I think unlike stop, there's no good reason for start to be called multiple times

drcrallen: Maybe? Prior behavior did not throw an error, which is why I did not do so here.

gianm: I feel like it makes sense to either not change the behavior, or to have it throw an error; silently skipping over double starts seems sketchy. At least the old behavior would probably blow up in some way and let us know something was being done wrong.

drcrallen: Fixed

himanshug:
:+1:

drcrallen:
Bouncing for travis

xvrl:
@drcrallen rebase on current master, it will help with travis

drcrallen:
ding! winner

PR #32, opened 2015-10-01
make ConcatSequence.toYielder(..) read next sequence only when

himanshug:
previous sequence has been iterated fully.

### DIFF
--- a/src/test/java/com/metamx/common/guava/ConcatSequenceTest.java
+++ b/src/test/java/com/metamx/common/guava/ConcatSequenceTest.java
@@ -17,15 +17,18 @@
 package com.metamx.common.guava;
 
 import com.google.common.base.Function;
+import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import junit.framework.Assert;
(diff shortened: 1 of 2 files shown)


src/test/java/com/metamx/common/guava/ConcatSequenceTest.java
```diff
@@ -187,6 +190,85 @@ public void cleanup(Iterator<Sequence<Integer>> iterFromMake)
...
+              public Sequence<Integer> apply(final ImmutableList<Integer> input)
+              {
+                if (lastSeqFullyRead.getAndSet(false)) {
+                  return Sequences.simple(
+                      new Iterable<Integer>()
+                      {
+                        private Iterator<Integer> baseIter = input.iterator();
+
+                        @Override
+                        public Iterator<Integer> iterator()
```
drcrallen: You're coding in java8 aren't you :) this is missing some stuff that's causing the unit tests to fail in java7

himanshug: yes :) , just added remove method

drcrallen:
:+1:

PR #35, opened 2015-11-06
Parser interface docs; deprecate setFieldNames/getFieldNames.

gianm:
- Docs for Parser
- Deprecated setFieldNames/getFieldNames. They don't appear to be used in meaningful ways anywhere in Druid, and I think are just confusing things by being on the interface.

### DIFF
--- a/src/main/java/com/metamx/common/parsers/Parser.java
+++ b/src/main/java/com/metamx/common/parsers/Parser.java
@@ -20,9 +20,30 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
+/**
+ * Class that can parse Strings into Maps.
+ */
 public interface Parser<K, V>
 {
-  public Map<K, V> parse(String input) ;
-  public void setFieldNames(Iterable<String> fieldNames) ;
+  /**
+   * Parse a String into a Map.
+   *
+   * @throws ParseException if the String cannot be parsed
+   */
+  public Map<K, V> parse(String input);
+
+  /**
+   * Set the fieldNames that you expect to see in parsed Maps. Deprecated; Parsers should not, in general, be
+   * expected to know what fields they will return. Some individual types of parsers do need to know (like a TSV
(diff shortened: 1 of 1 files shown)


himanshug:
:+1:

src/main/java/com/metamx/common/parsers/Parser.java
```diff
@@ -20,9 +20,30 @@
...
+   */
+  @Deprecated
+  public void setFieldNames(Iterable<String> fieldNames);
+
+  /**
+   * Returns the fieldNames that we expect to see in parsed Maps, if known, or null otherwise. Deprecated; Parsers
+   * should not, in general, be expected to know what fields they will return.
+   */
+  @Deprecated
   public List<String> getFieldNames();
```
himanshug: actually, thinking more about it, it seems support for fieldNames is useful for csv, tsv and the regex, javascript parsers in #36 unlike the JSONParser and they shouldn't be deprecated.

gianm: @himanshug the thought behind deprecating these was:

- external usages of setFieldNames are generally immediately after construction of a specific parser class. So those usages don't really need to use a setter- they could instead use a constructor that offers fieldNames (those constructors already exist but are not always used).

- these methods are actually pretty nonsensical for the JsonPath parser (#34), because its FieldSpecs serve the same purpose, but they do it in a richer way that is specific to JsonPath.

So, I don't think these methods are really useful, nor do they generalize well to all parsers.

himanshug: ok, sounds reasonable, I agree field names do not make sense for json parsers but only for csv, tsv, regex parsers who can support those via specific constructors.

himanshug:
:+1:

@xvrl @nishantmonu51 @drcrallen  can one of you pls merge this?

fjy:
:+1:

PR #39, opened 2016-02-17
Make the exception shortcuts use safe string formatting

drcrallen:


src/main/java/com/metamx/common/StringUtils.java
```diff
@@ -66,4 +67,21 @@ public static String fromUtf8(final ByteBuffer buffer)
...
     }
   }
+
+  public static String safeFormat(String message, Object... formatArgs)
+  {
+    if(formatArgs == null || formatArgs.length == 0) {
+      return message;
+    }
+    try {
+      return String.format(message, formatArgs);
```
gianm: could still throw NPE, but that's probably fine as that is a really strange error for a function that usually has literals passed into it.

src/main/java/com/metamx/common/StringUtils.java
```diff
@@ -66,4 +67,21 @@ public static String fromUtf8(final ByteBuffer buffer)
...
+    }
+    catch (MissingFormatArgumentException e) {
```
gianm: `IllegalFormatException` would catch a lot more things

gianm: oh, I see you copied this from Logger. I still think the change is worth it though.

nishantmonu51:
:+1:

src/main/java/com/metamx/common/StringUtils.java
drcrallen: As you noted below, this is a direct refactor out of Logger

Suggest not changing this method in this PR since it is a direct move.

gianm: ok, that's fair. I think switching the type would be in the spirit of what safeFormat is intended to do though.

gianm:
:+1: with/without exception type change

PR #41, opened 2016-02-17
Add slf4j-simple to test scope optionally

drcrallen:
Without this tests complain as follows:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,6 +137,13 @@
             <version>3.0</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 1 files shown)


nishantmonu51:
:+1:

PR #44, opened 2016-03-09
Add @Deprecated catch for logging assuming slf4j convention

drcrallen:
It is very easy for people unfamiliar with the logging conventions used here to swap these arguments, sometimes not yielding the desired information. This PR shows that methodology as deprecated and hopefully helps track down erroneous logging usages.

This might potentially change logging if someone had done the following:

`LOG.error("found an error: %s", someException);`

The fix would be to do something like

`LOG.error("found an error: %s", someException.getMessage());`

### DIFF
--- a/src/test/java/com/metamx/common/logger/LoggerTest.java
+++ b/src/test/java/com/metamx/common/logger/LoggerTest.java
@@ -26,4 +26,14 @@ public class LoggerTest
     final Logger log = new Logger(LoggerTest.class);
     log.warn(message);
   }
+
+  @Test
+  public void testLegacyLogging()
+  {
+    final Logger log = new Logger(LoggerTest.class);
+    final Throwable throwable = new Throwable();
+    // These should show up in an IDE as deprecated, but shouldn't actually fail.
+    log.error("foo", throwable);
+    log.warn("foo", throwable);
+  }
 }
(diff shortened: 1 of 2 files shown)


src/test/java/com/metamx/common/logger/LoggerTest.java
```diff
@@ -26,4 +26,13 @@ public void testLogWithCrazyMessages()
...
   }
+
+  @Test
+  public void testLegacyLogging()
+  {
+    final Logger log = new Logger(LoggerTest.class);
+    final Throwable throwable = new Throwable();
+    log.error("foo", throwable);
+    log.warn("foo", throwable);
+  }
```
himanshug: i think you are expecting IDE to flag these as deprecated. but since there are no asserts, its difficult for someone to understand what is being "tested" here. can you add comment telling why this test needs to exist or may be remove it since it can't really "fail" whether or not the thing you expected works or not.

nishantmonu51:
:+1:

src/test/java/com/metamx/common/logger/LoggerTest.java
drcrallen: ok sure

drcrallen: Fixed

drcrallen:
@himanshug any other comments?

leventov:
+1. Don't see a better way to protect from this or exclude this situation.

leventov:
However seems that it uses inconsistent formatting.. { no the same line.

PR #45, opened 2016-03-16
Utility method for length estimation of utf8

navis:
From https://github.com/druid-io/druid/pull/2661

### DIFF
--- a/src/main/java/com/metamx/common/StringUtils.java
+++ b/src/main/java/com/metamx/common/StringUtils.java
@@ -25,7 +25,7 @@ import java.nio.charset.Charset;
 import java.util.IllegalFormatException;
 
 /**
- * As of right now (Dec 2014) the JVM is optimized around String charset variablse instead of Charset passing.
+ * As of right now (Dec 2014) the JVM is optimized around String charset variable instead of Charset passing.
  */
 public class StringUtils
 {
@@ -70,7 +70,7 @@ public class StringUtils
 
   public static String safeFormat(String message, Object... formatArgs)
   {
-    if(formatArgs == null || formatArgs.length == 0) {
+    if (formatArgs == null || formatArgs.length == 0) {
       return message;
     }
     try {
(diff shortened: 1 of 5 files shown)


src/main/java/com/metamx/common/StringUtils.java
```diff
@@ -84,4 +84,24 @@ public static String safeFormat(String message, Object... formatArgs)
       return bob.toString();
     }
   }
+
+  // should be used only for estimation.. does not check validity of format
```
drcrallen: Can you add javadoc that this is usually equivalent to `StringUtils.toUtf8(value).length`?

drcrallen:
@navis there's a lot of fixes in here, can you either split by commit or split them into different PRs?

src/main/java/com/metamx/common/StringUtils.java
xvrl: it would also be good to document in the cases in which it is not equal to `StringUtils.toUtf8(value).length`. Does it tend to under-estimate or does it tend to over-estimate?

src/test/java/com/metamx/common/StringUtilsTest.java
```diff
@@ -26,6 +26,34 @@
...
+  };
+
+  @Test
+  public void binaryLengthAsUTF8Test() throws UnsupportedEncodingException
+  {
+    for (String string : TEST_STRINGS) {
+      Assert.assertEquals(StringUtils.toUtf8(string).length, StringUtils.binaryLengthAsUTF8(string));
+    }
+  }
+
```
xvrl: let's also add a tests where the length are not equal to clarify that this is the expected behavior

src/main/java/com/metamx/common/StringUtils.java
navis: Basically the same but it does not validate whether the string has valid UTF format. For invalid chars, this method over-estimates the length(length of invalid char = 1 for encoding).

src/test/java/com/metamx/common/StringUtilsTest.java
navis: ok. I'll add invalid char cases.

navis:
It's merged into druid

PR #46, opened 2016-04-07
Use file channel in smoosher to minimize copy overheads

navis:
Current `write(ByteBuffer buffer)` implementation of smoother is something like this,
```
Channels.newChannel(out).write(buffer)
```

And agagin,
```
public static ReadableByteChannel newChannel(final InputStream in) {
        if (in instanceof FileInputStream &&
            FileInputStream.class.equals(in.getClass())) {
            return ((FileInputStream)in).getChannel();
        }
        return new ReadableByteChannelImpl(in);
    }
```

but because `out` is a BufferedOuputStream, this call always makes wrapper which copies ByteBuffer handed over to byte[], negating good intention of using it. I think it would be better to use FileChannel directly.

### DIFF
--- a/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
+++ b/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
@@ -40,6 +40,7 @@ import java.io.Writer;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.FileChannel;
+import java.nio.channels.GatheringByteChannel;
 import java.nio.channels.WritableByteChannel;
 import java.util.Arrays;
 import java.util.List;
@@ -160,7 +161,19 @@ public class FileSmoosher implements Closeable
         return verifySize(currOut.write(in));
       }
 
-      private int verifySize(int bytesWrittenInChunk) throws IOException
+      @Override
+      public long write(ByteBuffer[] srcs, int offset, int length) throws IOException
+      {
+        return verifySize(currOut.write(srcs, offset, length));
+      }
+
+      @Override
+      public long write(ByteBuffer[] srcs) throws IOException
+      {
+        return verifySize(currOut.write(srcs));
+      }
+
+      private int verifySize(long bytesWrittenInChunk) throws IOException
       {
         bytesWritten += bytesWrittenInChunk;
 
@@ -171,7 +184,7 @@ public class FileSmoosher implements Closeable
           throw new ISE("Wrote[%,d] bytes for something of size[%,d].  Liar!!!", bytesWritten, size);
         }
 
-        return bytesWrittenInChunk;
+        return Ints.checkedCast(bytesWrittenInChunk);
       }
 
       @Override
@@ -247,8 +260,8 @@ public class FileSmoosher implements Closeable
   public static class Outer implements SmooshedWriter
   {
     private final int fileNum;
-    private final WritableByteChannel channel;
     private final int maxLength;
+    private final GatheringByteChannel channel;
 
     private int currOffset = 0;
 
@@ -283,10 +296,22 @@ public class FileSmoosher implements Closeable
     @Override
     public int write(InputStream in) throws IOException
     {
-      return addToOffset(Ints.checkedCast(ByteStreams.copy(Channels.newChannel(in), channel)));
+      return addToOffset(ByteStreams.copy(Channels.newChannel(in), channel));
     }
 
-    public int addToOffset(int numBytesWritten)
+    @Override
+    public long write(ByteBuffer[] srcs, int offset, int length) throws IOException
+    {
+      return addToOffset(channel.write(srcs, offset, length));
+    }
+
+    @Override
+    public long write(ByteBuffer[] srcs) throws IOException
+    {
+      return addToOffset(channel.write(srcs));
+    }
+
+    public int addToOffset(long numBytesWritten)
     {
       if (numBytesWritten > bytesLeft()) {
         throw new ISE("Wrote more bytes[%,d] than available[%,d]. Don't do that.", numBytesWritten, bytesLeft());
@@ -294,7 +319,7 @@ public class FileSmoosher implements Closeable
 
       currOffset += numBytesWritten;
 
-      return numBytesWritten;
+      return Ints.checkedCast(numBytesWritten);
     }
 
     @Override
@@ -311,7 +336,7 @@ public class FileSmoosher implements Closeable
   }
 
   private static class BufferedWritableByteChannel
-      implements WritableByteChannel
+      implements GatheringByteChannel
   {
     private static final int MAX_BUFFER_SIZE = 65536;
     private static final int DEFAULT_BUFFER_SIZE = 8192;  // default of buffered output stream
@@ -347,6 +372,22 @@ public class FileSmoosher implements Closeable
         }
       }
       return src.position() - position;
+    }
+
+    @Override
+    public long write(ByteBuffer[] srcs, int offset, int length) throws IOException
+    {
+      long written = 0;
+      for (int i = offset; i < length; i++) {
+        written += write(srcs[i]);
+      }
+      return written;
+    }
+
+    @Override
+    public long write(ByteBuffer[] srcs) throws IOException
+    {
+      return write(srcs, 0, srcs.length);
     }
 
     private void flush() throws IOException
--- a/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
+++ b/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
@@ -28,7 +28,6 @@ import com.metamx.common.IAE;
 import com.metamx.common.ISE;
 import com.metamx.common.MappedByteBufferHandler;
 
-import java.io.BufferedOutputStream;
 import java.io.BufferedWriter;
 import java.io.Closeable;
 import java.io.File;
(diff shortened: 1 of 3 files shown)


src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -132,7 +132,7 @@ public void testBehaviorWhenReportedSizesSmall() throws Exception
       Assert.assertTrue(exceptionThrown);
       File[] files = baseDir.listFiles();
       Assert.assertEquals(1, files.length);
-      Assert.assertEquals(0, files[0].length());
+      Assert.assertEquals(4, files[0].length());
```
xvrl: @navis any reason this is changing?

navis: it was buffered stream before but is not, making that difference. I've implemented buffered channel and it's restored as before.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -299,38 +309,130 @@ public int bytesLeft()
...
+      return channel.isOpen();
     }
 
     @Override
     public void close() throws IOException
     {
-      open = false;
-      out.close();
+      if (buffer.position() != 0) {
+        flush();
```
drcrallen: if flush throws IOException, does close() still need called?

navis: hm.. good.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -299,38 +309,137 @@ public int bytesLeft()
...
     {
-      WritableByteChannel channel = Channels.newChannel(out);
       return addToOffset(channel.write(buffer));
     }
 
     @Override
     public int write(InputStream in) throws IOException
     {
-      return addToOffset(Ints.checkedCast(ByteStreams.copy(in, out)));
+      return addToOffset(ByteStreams.copy(Channels.newChannel(in), channel));
```
leventov: Double buffering here. `Channels.newChannel()` returns a buffered channel, and then `ByteStreams.copy()` uses a buffer internally. Made https://github.com/google/guava/issues/2559 about this but it's faster and easier to fix internally in the meantime

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -268,16 +279,15 @@ static File makeChunkFile(File baseDir, int i)
...
     private final int fileNum;
-    private final OutputStream out;
     private final int maxLength;
+    private final GatheringByteChannel channel;
 
-    private boolean open = true;
     private int currOffset = 0;
 
-    Outer(int fileNum, OutputStream out, int maxLength)
+    Outer(int fileNum, FileChannel channel, int maxLength)
```
leventov: Don't see the point of having two separate SmoothedWriter impls: Outer and here: https://github.com/metamx/java-util/pull/46/files#diff-99d3b3a2a7085d596b82e8831f793554R163. Why this is not just a single class, to remove boilerplate?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -268,16 +279,15 @@ static File makeChunkFile(File baseDir, int i)
...
     {
       this.fileNum = fileNum;
-      this.out = out;
+      this.channel = new BufferedWritableByteChannel(channel);
```
leventov: Have this been tested (benchmarked), that extra buffering is required here? I doubt it is.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -252,7 +263,7 @@ private Outer getNewCurrOut() throws FileNotFoundException
     final int fileNum = outFiles.size();
     File outFile = makeChunkFile(baseDir, fileNum);
     outFiles.add(outFile);
-    return new Outer(fileNum, new BufferedOutputStream(new FileOutputStream(outFile)), maxChunkSize);
+    return new Outer(fileNum, new FileOutputStream(outFile).getChannel(), maxChunkSize);
```
leventov: I think FileOutputStream should be created with try-with-resources, just in case.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
navis: Yes, I've neglected this because this is not used in Druid. But seemed good this to be fixed.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
navis: Yep, I wish that too, but it was like that before. I thought it's possible once but also wanted minimize the changes.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
navis: Not-buffering (was original patch) makes a test-case fail. (see https://github.com/metamx/java-util/pull/46#discussion_r58923324)

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
navis: good

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: How does it makes any difference? Buffering shouldn't change semantics of streams/channels, and therefore the code that use them. Maybe something was broken/misused?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: In the meantime I was experimenting with this class, try-with-resources couldn't be used right here, because the outer is returned, and TWR will close channel. So I think it should be like this:
 1. Pass FOS into Outer instead of Channel, and save it as a field in Outer.
 2. Extract, (optionally wrap with Buffered..Channel) and store Channel in a field in Outer in the Outer's constructor.
 3. In addition to channel, close FOS in Outer's close.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: Later I realised the purpose of this separation, I'm ok with it now

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: Agree, could be a separate PR. Also found this idiom (`ByteStream.copy(Channels.newChannel(), ...`) is used in Druid in several places, so it's more a Druid's issue than java-util's

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -299,38 +309,137 @@ public int bytesLeft()
...
-      return open;
+      return channel.isOpen();
     }
 
     @Override
     public void close() throws IOException
     {
-      open = false;
-      out.close();
+      try {
```
leventov: I suggest to use TWR "to close" idiom here as more concise and benefitting Throwable suppression rather than losing them:
```
try (Closeable toClose = channel) {
  if (buffer.position() != 0) {
    flush(true);
  }
}
```

leventov:
@navis do you have time to address the comments? I can do this myself and make a different PR based on your PR, but if it is merged with squash your authorship will be lost.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
navis: @leventov Sorry for delay. The test throws exception and does not flush or close underlying stream, making difference in file size. If it's acceptable change by metamx, I'm really happy on removing buffering channel.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: @navis looking at the modern OpenJDK implementation, FOS additionally closes FileDescriptor. I'm not sure if Channel closes it too or not, but anyway it's right to close an "outer" object (ie fos, because channel is obtained from fos in this case), because 1) in different (maybe future) version of OpenJDK, or in a different JDK this may be different, and fos closes something non trivial in addition to what channel closes.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
navis: done

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: @navis if you don't want to benchmark druid's code that uses this (what I understand, because it's tedious), I'm pro removing buffering, because it will allow to remove the hand-written buffering channel class. We can return to this only if we will see perf regression in druid, when we try to upgrade it to a version of java-util, which has this change.

leventov:
@navis could you please make two things?

 1. Remove channel buffering: it causes `deterministicFileUnmapping()` test to fail now, because it 1) allocates a direct buffer 2) duplicates mapped buffers.

 2. Close  FOS in Outer as suggested here: https://github.com/metamx/java-util/pull/46#r78347977

leventov:
And after those changes `SmooshedFileMapperTest.testBehaviorWhenReportedSizesSmall()` will fail. The last assertion in this test,  namely `Assert.assertEquals(0, files[0].length());` should simply be removed because it seems to rely on `FileSmoosher` implementation details. Having or not some bytes written to the last smooshed file after `ISE` is thrown doesn't seem to be an important difference, because metadata is not written anyway and this smooshed dir couldn't be read later.

leventov:
@navis do you have capacity for doing this?

navis:
@leventov Addressed comments

leventov:
@navis Thanks!

PR #50, opened 2016-09-02
Unmap files to add eagerly in FileSmoother

leventov:
The exact purpose of this is to enable running Druid's IndexMergeBenchmark in Windows, but this is also universally healthy becauseit releases memory determenistically, otherwise files are unmapped when MappedByteBuffers are garbage-collected

### DIFF
--- a/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
+++ b/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
@@ -25,6 +25,7 @@ import com.google.common.io.ByteStreams;
 import com.google.common.io.Closeables;
 import com.google.common.io.Files;
 import com.google.common.primitives.Ints;
+import com.metamx.common.ByteBufferUtils;
 import com.metamx.common.IAE;
 import com.metamx.common.ISE;
 
@@ -40,6 +41,7 @@ import java.io.OutputStream;
 import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.ByteBuffer;
+import java.nio.MappedByteBuffer;
 import java.nio.channels.Channels;
 import java.nio.channels.WritableByteChannel;
 import java.util.Arrays;
@@ -53,6 +55,7 @@ import java.util.Set;
  * <p/>
  * It does not split input files among separate output files, instead the various "chunk" files will
  * be varying sizes and it is not possible to add a file of size greater than Integer.MAX_VALUE
+ * TODO(leventov): consider making this class final
  */
 public class FileSmoosher implements Closeable
 {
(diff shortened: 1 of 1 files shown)


drcrallen:
Looks good but needs to have verification that tests still pass in druid on linux.

drcrallen:
Wait a sec, I need to double-check this.

drcrallen:
There are explicit tests in druid about properly unmapping files. It would be good to make sure this still functions in those tests.

leventov:
@drcrallen could you please point to them? This commit in java-util is only a part of the change. I'm also going to propose some fixes to druid itself, regarding specifically leaking file mappings. So those tests apparently don't catch the problems and should be updated too..

drcrallen:
@leventov can you peek at https://github.com/metamx/java-util/pull/46 by @navis and see how much it collides with this one? I completely forgot #46 existed.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -105,12 +108,17 @@ private FileSmoosher(
...
   }
 
   public void add(String name, File fileToAdd) throws IOException
   {
-    add(name, Files.map(fileToAdd));
+    MappedByteBuffer mappedFileBuffer = Files.map(fileToAdd);
+    try {
+      add(name, mappedFileBuffer);
+    } finally {
+      ByteBufferUtils.unmap(mappedFileBuffer);
```
gianm: try-with-resources preferable here to avoid exception chomping.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -53,6 +55,7 @@
  * <p/>
  * It does not split input files among separate output files, instead the various "chunk" files will
  * be varying sizes and it is not possible to add a file of size greater than Integer.MAX_VALUE
+ * TODO(leventov): consider making this class final
```
gianm: I know this isn't Druid, but generally in Druid we try to minimize TODOs in the code, in favor of Github issues or in favor of just actually doing the thing if it's really important.

leventov: @gianm thanks, will know.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
leventov: Yes... also there is #46, and also there is an idea to approach this more fundamentially with ResourceHandlers, discussed in https://github.com/druid-io/druid/pull/3422. Another option is to move away from ByteBuffers, to e. g. Netty's, Aeron's or Chronicle's alternatives, which are better in many ways. So I close this PR for now anyway.

PR #51, opened 2016-09-12
Add ResourceHandler, FileUtils.map() returning ResourceHandler

leventov:
And improve resource closing in FileSmoosher and SmooshedFileMapper

Better version of #50


src/main/java/com/metamx/common/ResourceHandler.java
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * Returns the wrapped resource.
+   */
+  T get();
+
+  /**
+   * Closes the wrapped resource.
+   */
+  @Override
+  void close();
```
drcrallen: `AutoCloseable.close()` allows `Exception`, is there a need to erase that signature here?

src/main/java/com/metamx/common/FileUtils.java
```diff
@@ -125,4 +129,54 @@ public void addFile(File file)
...
+   *
+   * @see FileChannel#map(FileChannel.MapMode, long, long)
+   */
+  public static ResourceHandler<MappedByteBuffer> map(File file) throws IOException
+  {
+    MappedByteBuffer mappedByteBuffer = Files.map(file);
+    return new MappedByteBufferHandler(mappedByteBuffer);
+  }
+
+  private static class MappedByteBufferHandler implements ResourceHandler<MappedByteBuffer> {
```
drcrallen: Does this need to have `<MappedByteBuffer>` instead of just `<ByteBuffer>` as the signature?

drcrallen:
This might collide a bit with `io.druid.collections.ResourceHolder`, if you have any thoughts on how to keep too many "resource holding interfaces" from polluting the namespace please let them be known.

src/main/java/com/metamx/common/ResourceHandler.java
leventov: This interface is copied from Druid: `io.druid.collections.ResourceHandler`. It erases exception, apparently because no present application requires throwing exception, but having it in signature it more noisy on usage side

drcrallen:
Overall this is a very reasonable, appropriate, and well executed patch. I would like to see some way to limit the quantity of resource handling interfaces.

One concern is that `io.druid.collections.ResourceHolder` extends `Closeable` whereas the proposed addition here extends the more restrictive `Autocloseable`, which means if `io.druid.collections.ResourceHolder` were to extend the proposed interface, it would also need to extend `Closeable`  *if* we were to ensure zero interface changes. We can always change the interface though, so that is by no means a strict requirement.

src/main/java/com/metamx/common/FileUtils.java
leventov: `Files.map()` returns `MappedByteBuffer`. I don't see how returning `ByteBuffer` is more agile, because I don't see why this MBB is ever going to be wrapped/sliced before wrapping with ResourceHandler.

leventov: You mean, a top-level package-private class in the same file? I find it irritating in IntelliJ because need to wrap/unwrap by clicking on the icon in project explorer to navigate to a class within a single file.

leventov:
It intentionally shadows Druid's ResourceHandler, because 1) ResourceHandler finds it's usage in java-utils source itself 2) Best place for map() is FileUtils, there is no similar class or even package in Druid codebase.

Druid's ResourceHandler is used only internally, so I don't see a big problem if they co-exist with java-utils, or Druid's could be deprecated, or removed.

Another way is adding `io.druid.collections.ResourceHandler` to java-util directly but it would be strange and module technologies like OSGi or Java 9 modules are not sympathetic to things like this (when multiple modules/project share packages).

leventov:
> @leventov please add tests to show the functionality added here is accomplishing what it is intended to accomplish.

The direct reason for this is still ability to run some druid's benchmarks in windows. So it's impossible to write a sensible test to this to run in Linux.

Also some Druid's `Files.map()`-using code could be simplified, https://github.com/druid-io/druid/pull/3422#issuecomment-245358665

leventov:
> One concern is that io.druid.collections.ResourceHolder extends Closeable whereas the proposed addition here extends the more restrictive Autocloseable, which means if io.druid.collections.ResourceHolder were to extend the proposed interface, it would also need to extend Closeable if we were to ensure zero interface changes. We can always change the interface though, so that is by no means a strict requirement.

I didn't mean `io.druid.collections.ResourceHolder` to extend the added `ResourceHandler`, I see no point by saving literally a couple of lines of code. `io.druid.collections.ResourceHolder` could be removed if used only internally, or if there is a suspection that some extensions use this interface, we can left it in the codebase until the next major Druid release, as is.

leventov:
Or, the added ResourceHolder could also be made to extend `Closeable`, it's ok.

drcrallen:
@leventov it's not about saving code, its about saving developer sanity. If someone sees there are two interfaces in place that do essentially the same thing, and have no logical reason they should be different, then that muddles the waters more than it should.

I propose having ResourceHandler from java-util be the interface used in druid instead of ResourceHolder.

leventov:
@gianm could you please review and/or express your opinion as Druid commiter about replacing `io.druid.collections.ResourceHandler` with `com.metamx.common.ResourceHandler`?

drcrallen:
As per https://groups.google.com/d/msg/druid-development/SQK4FYxzeXw/i2jx_atBAAAJ looks like the interface is staying here

leventov:
@drcrallen as pulling generic `ResourceHandler` into java-util aimed to replace Druid's `ResourceHandler`, what is not going to happen, I can specialize it, i. e. have just `MappedByteBufferHandler` in java-util. Makes sense?

src/main/java/com/metamx/common/FileUtils.java
drcrallen: Looking at potential future uses, if something were to need a `ResourceHandler<ByteBuffer>` they would either have to `ResourceHandler<? extends ByteBuffer>` or do some strange casting. Correct me if I'm wrong but I don't think

```java
final ResourceHandler<ByteBuffer> handler = FileUtils.map(file);
```

would work in such a case. 

```java
final ResourceHandler<? extends ByteBuffer> handler = FileUtils.map(file);
```

should work.

I just tried it out and

```java
Collection<? extends Collection<? extends Number>> test = ImmutableList.of(ImmutableList.<Double>of(1.0), ImmutableList.<Long>of(3L));
```
parses correctly but

```java
Collection<? extends Collection<? extends Number>> test = new ArrayList<>();
test.add(ImmutableList.<Double>of(1.0));
``` 

does not. Wildcards in java can be weird sometimes.

Having the public interface (`get(File)`) return the least common denominator of what is needed helps prevent such scenarios.

drcrallen: As extra fun

```java
Collection<Collection<? extends Number>> test = new ArrayList<>();
test.add(ImmutableList.<Double>of(1.0));
```
does work

drcrallen: Long story short, I prefer least common denominator type parameters, but if you still want `MappedByteBuffer` after hearing the info above then I'm not concerned about it until an issue arrises in code we can point at and say "It doesn't work in this scenario"

src/main/java/com/metamx/common/io/smoosh/SmooshedFileMapper.java
```diff
@@ -131,10 +132,20 @@ public ByteBuffer mapFile(String name) throws IOException
...
+        ByteBufferUtils.unmap(mappedByteBuffer);
+      } catch (Throwable t) {
+        if (thrown == null) {
+          thrown = t;
+        } else {
+          thrown.addSuppressed(t);
+        }
+      }
     }
+    Throwables.propagateIfPossible(thrown);
```
drcrallen: This changes behavior as it will now completely swallow `IOException` instead of propagating it, or even logging it.

Is there a reason the exception should be swallowed now?

src/main/java/com/metamx/common/ResourceHandler.java
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+
+package com.metamx.common;
+
+/**
+ * Facilitates using try-with-resources with resources that don't implement {@link AutoCloseable}, e. g.
+ * {@link java.nio.ByteBuffer}s.
+ *
+ * <p>This interface replaces {@code io.druid.collections.ResourceHandler}.
```
drcrallen: don't forget to remove, you can say its modeled after, but not replaces as per discussion.

src/main/java/com/metamx/common/io/smoosh/SmooshedFileMapper.java
leventov: `IOException` was not possible here, so I just removed unneeded `throws IOException`. Only `RuntimeException` or `Error` are possible. So `Throwables.propagateIfPossible(thrown);` is just a way to rethrow, without javac complaining.

src/main/java/com/metamx/common/io/smoosh/SmooshedFileMapper.java
```diff
@@ -131,10 +132,20 @@ public ByteBuffer mapFile(String name) throws IOException
...
   @Override
-  public void close() throws IOException
+  public void close()
   {
+    Throwable thrown = null;
     for (MappedByteBuffer mappedByteBuffer : buffersList) {
-      ByteBufferUtils.unmap(mappedByteBuffer);
+      try {
+        ByteBufferUtils.unmap(mappedByteBuffer);
+      } catch (Throwable t) {
```
drcrallen: FYI, there are a few `Throwable`s that, in general, should probably not allow continuation in the running process.

http://www.scala-lang.org/api/2.10.6/index.html#scala.util.control.NonFatal$ has some good examples of things which should probably not continue. 

Digging through the call stack for `unmap` though, it doesn't look like it throws anything that should be caught.

src/main/java/com/metamx/common/io/smoosh/SmooshedFileMapper.java
drcrallen: Gotcha

src/main/java/com/metamx/common/io/smoosh/SmooshedFileMapper.java
drcrallen: In this particular case I think the behavior is OK due to the very limited nature of what is accomplished, or intended, by the unmap.

leventov:
@cheddar @drcrallen could you please express your opinion on the current state of the PR using the new Github's review system?

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -35,11 +40,13 @@
   public void testSanity() throws Exception
   {
     File baseDir = Files.createTempDir();
+    baseDir.deleteOnExit();
```
drcrallen: Can you use the `TemporaryFolder` `@Rule` for junit for this?

leventov:
@drcrallen could you please review latest changes?

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -18,126 +18,134 @@
...
-        Assert.assertTrue(e.getMessage().contains("Liar!!!"));
-        exceptionThrown = true;
-      }
+    File baseDir = folder.newFolder("base");
 
-      Assert.assertTrue(exceptionThrown);
-      File[] files = baseDir.listFiles();
-      Assert.assertEquals(1, files.length);
-      Assert.assertEquals(0, files[0].length());
+    FileSmoosher smoosher = new FileSmoosher(baseDir, 21);
```
drcrallen: Should this one be handled in a try-with-resources?

leventov: fixed

PR #52, opened 2016-09-14
Remove reflection in ByteBufferUtils

drcrallen:
* Also adds test


drcrallen:
@leventov I don't think the `ByteBufferUtils` needs to be as complex

@cheddar would you mind taking a peek at this?

leventov:
Reflection approach is used, when somebody needs graceful fail when non-HotSpot JVM is used. I. e. somebody trying to usejava-util from within Websphere will be in thouble, because it runs on IBM's J9. If we explicitly support only Hotspot JVM I think this is a good change.

leventov:
Ok, i see that exceptions were thrown in static blocks before, so It was not compatible with other JVMs anyway.

src/main/java/com/metamx/common/ByteBufferUtils.java
```diff
@@ -17,55 +17,29 @@
...
+  public static void free(ByteBuffer buffer)
+  {
     try {
-      clean.invoke(getCleaner.invoke(buffer));
-    } catch(IllegalAccessException | InvocationTargetException e) {
+      if (buffer.isDirect()) {
+        clean((DirectBuffer) buffer);
+      }
+    }
+    catch (Exception e) {
```
leventov: Is try-catch needed here?

src/test/java/com/metamx/common/ByteBufferUtilsTest.java
```diff
@@ -0,0 +1,67 @@
...
+  @Test
+  public void testUnmapDoesntCrashJVM() throws Exception
+  {
+    final File file = temporaryFolder.newFile("some_mmap_file");
+    try (final OutputStream os = new BufferedOutputStream(new FileOutputStream(file))) {
+      final byte[] data = new byte[4096];
+      Arrays.fill(data, (byte) 0x5A);
+      os.write(data);
+    }
+    final MappedByteBuffer mappedByteBuffer;
```
leventov: Why not just `mappedByteBuffer = Files.map(file)`?

drcrallen: Because I forgot it existed, fixed!

src/main/java/com/metamx/common/ByteBufferUtils.java
drcrallen: Don't think so. Fixed, er... removed

PR #55, opened 2016-09-26
File smoosher changes

akashdw:
This pr is to support `FileSmoosher` changes proposed in https://github.com/druid-io/druid/issues/3513 .

`FileSmoosher` detect when it already has an OutputStream open and redirect newly opened OutputStreams to new files on the file system. When any of the open OutputStream objects are closed, they will also check to see if any of the other files have been closed in the meantime. If there are OutputStreams that have been closed, they will be copied on to the main smoosh file and the extra underlying file will be cleaned up. This has the downside of introducing an extra copy to these files, but it given that the strategy will only be used when absolutely necessary, this shouldn’t result in any noticeable performance degradation during indexing.

### DIFF
--- a/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
+++ b/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
@@ -66,15 +66,15 @@ public class FileSmoosher implements Closeable
 
   private final List<File> outFiles = Lists.newArrayList();
   private final Map<String, Metadata> internalFiles = Maps.newTreeMap();
-  private List<File> files = Lists.newArrayList();
-  private List<File> createFiles = Lists.newArrayList();
+  private List<File> completedFiles = Lists.newArrayList();
+  private List<File> filesInProcess = Lists.newArrayList();
 
   private Outer currOut = null;
   private boolean writerCurrentlyInUse = false;
 
   public FileSmoosher(
       File baseDir
-  )
+      )
   {
     this(baseDir, Integer.MAX_VALUE);
   }
@@ -82,7 +82,7 @@ public class FileSmoosher implements Closeable
   public FileSmoosher(
       File baseDir,
       int maxChunkSize
-  )
+      )
   {
     this.baseDir = baseDir;
     this.maxChunkSize = maxChunkSize;
@@ -204,13 +204,10 @@ public class FileSmoosher implements Closeable
         if (bytesWritten != size) {
           throw new IOException(
               String.format("Expected [%,d] bytes, only saw [%,d], potential corruption?", size, bytesWritten)
-              );
+          );
         }
 
-        if(!writerCurrentlyInUse)
-        {
-          mergeWithSmoosher();
-        }
+        mergeWithSmoosher();
       }
     };
   }
@@ -218,12 +215,11 @@ public class FileSmoosher implements Closeable
   private void mergeWithSmoosher() throws IOException
   {
     //get processed elements from the stack and write.
-    List<File> fileToProcess = new ArrayList<>(files);
-    files = Lists.newArrayList();
+    List<File> fileToProcess = new ArrayList<>(completedFiles);
+    completedFiles = Lists.newArrayList();
     for(File file: fileToProcess)
     {
       add(file);
-      createFiles.remove(file);
       file.delete();
     }
   }
@@ -231,7 +227,7 @@ public class FileSmoosher implements Closeable
   private SmooshedWriter delegateSmooshedWriter(final String name,final long size) throws IOException
   {	
     final File tmpFile = new File(baseDir, name);
-    createFiles.add(tmpFile);
+    filesInProcess.add(tmpFile);
     return new SmooshedWriter () {
       private int currOffset = 0;
       private boolean open = true;
@@ -242,7 +238,9 @@ public class FileSmoosher implements Closeable
       {
         open = false;
         out.close();
-        files.add(tmpFile);
+        completedFiles.add(tmpFile);
+        filesInProcess.remove(tmpFile);
+
         if(!writerCurrentlyInUse ) {
           mergeWithSmoosher();
         }
@@ -255,26 +253,26 @@ public class FileSmoosher implements Closeable
       @Override
       public int write(ByteBuffer buffer) throws IOException
       {
-        FileChannel channel = out.getChannel();
+        WritableByteChannel channel = Channels.newChannel(out);
         int bytesWritten = channel.write(buffer);
-        return  addToOffset(bytesWritten);
+        addToOffset(bytesWritten);
+        return bytesWritten;
       }
 
       @Override
       public int write(InputStream in) throws IOException
       {
-        return addToOffset(Ints.checkedCast(java.nio.file.Files.copy(in, tmpFile.toPath())));
+        int bytesWritten = Ints.checkedCast(ByteStreams.copy(in, out));
+        addToOffset(bytesWritten);
+        return bytesWritten;
       }
 
-      public int addToOffset(int numBytesWritten)
+      private void addToOffset(int numBytesWritten)
       {
         if (numBytesWritten > bytesLeft()) {
           throw new ISE("Wrote more bytes[%,d] than available[%,d]. Don't do that.", numBytesWritten, bytesLeft());
         }
-
         currOffset += numBytesWritten;
-
-        return numBytesWritten;
       }
 
       @Override
@@ -289,15 +287,15 @@ public class FileSmoosher implements Closeable
   public void close() throws IOException
   {
     //book keeping checks on created file.
-    if(!createFiles.isEmpty())
+    if(!completedFiles.isEmpty())
     {
-      for(File file: createFiles)
+      for(File file: completedFiles)
       {
         file.delete();
       }
-      throw new ISE(String.format("%d writers needs to be closed before closing smoosher.", createFiles.size()));
+      throw new ISE(String.format("%d writers needs to be closed before closing smoosher.", filesInProcess.size()));
     }
-    
+
     if (currOut != null) {
       currOut.close();
     }
@@ -316,8 +314,8 @@ public class FileSmoosher implements Closeable
                 metadata.getFileNum(),
                 metadata.getStartOffset(),
                 metadata.getEndOffset()
-            )
-        );
+                )
+            );
         out.write("\n");
       }
     }
@@ -376,24 +374,25 @@ public class FileSmoosher implements Closeable
     public int write(ByteBuffer buffer) throws IOException
     {
       WritableByteChannel channel = Channels.newChannel(out);
-      return addToOffset(channel.write(buffer));
+      int bytesWritten = channel.write(buffer);
+      addToOffset(bytesWritten);
+      return bytesWritten;
     }
 
     @Override
     public int write(InputStream in) throws IOException
     {
-      return addToOffset(Ints.checkedCast(ByteStreams.copy(in, out)));
+      int bytesWritten = Ints.checkedCast(ByteStreams.copy(in, out));
+      addToOffset(bytesWritten);
+      return bytesWritten;
     }
 
-    public int addToOffset(int numBytesWritten)
+    private void addToOffset(int numBytesWritten)
     {
       if (numBytesWritten > bytesLeft()) {
         throw new ISE("Wrote more bytes[%,d] than available[%,d]. Don't do that.", numBytesWritten, bytesLeft());
       }
-
       currOffset += numBytesWritten;
-
-      return numBytesWritten;
     }
 
     @Override
--- a/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
+++ b/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
@@ -74,7 +74,7 @@ public class FileSmoosher implements Closeable
 
   public FileSmoosher(
       File baseDir
-      )
+  )
   {
     this(baseDir, Integer.MAX_VALUE);
   }
@@ -82,7 +82,7 @@ public class FileSmoosher implements Closeable
   public FileSmoosher(
       File baseDir,
       int maxChunkSize
-      )
+  )
   {
     this.baseDir = baseDir;
     this.maxChunkSize = maxChunkSize;
@@ -206,7 +206,6 @@ public class FileSmoosher implements Closeable
               String.format("Expected [%,d] bytes, only saw [%,d], potential corruption?", size, bytesWritten)
           );
         }
-
         mergeWithSmoosher();
       }
     };
@@ -314,8 +313,8 @@ public class FileSmoosher implements Closeable
                 metadata.getFileNum(),
                 metadata.getStartOffset(),
                 metadata.getEndOffset()
-                )
-            );
+            )
+        );
         out.write("\n");
       }
     }
--- a/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
+++ b/src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
@@ -42,7 +42,9 @@ import java.io.OutputStreamWriter;
 import java.io.Writer;
 import java.nio.ByteBuffer;
 import java.nio.channels.Channels;
+import java.nio.channels.FileChannel;
 import java.nio.channels.WritableByteChannel;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
@@ -65,8 +67,11 @@ public class FileSmoosher implements Closeable
 
   private final List<File> outFiles = Lists.newArrayList();
   private final Map<String, Metadata> internalFiles = Maps.newTreeMap();
+  private List<File> files = Lists.newArrayList();
+  private List<File> createFiles = Lists.newArrayList();
 
   private Outer currOut = null;
+  private boolean writerCurrentlyInUse = false;
 
   public FileSmoosher(
       File baseDir
@@ -148,6 +153,11 @@ public class FileSmoosher implements Closeable
 
   public SmooshedWriter addWithSmooshedWriter(final String name, final long size) throws IOException
   {
+    if(writerCurrentlyInUse)
+    {
+      return delegateSmooshedWriter (name , size);
+    }
+
     if (size > maxChunkSize) {
       throw new IAE("Asked to add buffers[%,d] larger than configured max[%,d]", size, maxChunkSize);
     }
@@ -160,7 +170,7 @@ public class FileSmoosher implements Closeable
     }
 
     final int startOffset = currOut.getCurrOffset();
-
+    writerCurrentlyInUse = true;
     return new SmooshedWriter()
     {
       private boolean open = true;
@@ -203,6 +213,7 @@ public class FileSmoosher implements Closeable
       {
         open = false;
         internalFiles.put(name, new Metadata(currOut.getFileNum(), startOffset, currOut.getCurrOffset()));
+        writerCurrentlyInUse = false;
 
         if (bytesWritten != currOut.getCurrOffset() - startOffset) {
           throw new ISE("WTF? Perhaps there is some concurrent modification going on?");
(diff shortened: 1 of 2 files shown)


src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -131,6 +136,11 @@ public void add(String name, List<ByteBuffer> bufferToAdd) throws IOException
 
   public SmooshedWriter addWithSmooshedWriter(final String name, final long size) throws IOException
   {
+    if(writerCurrentlyInUse)
+    {
+      return delegateSmooshedWriter (name , size);
```
cheddar: weird spacing

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -64,8 +66,11 @@
 
   private final List<File> outFiles = Lists.newArrayList();
   private final Map<String, Metadata> internalFiles = Maps.newTreeMap();
+  private List<File> files = Lists.newArrayList();
+  private List<File> createFiles = Lists.newArrayList();
```
cheddar: Should this be past-tense?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -186,22 +196,108 @@ public void close() throws IOException
...
+        writerCurrentlyInUse = false;
 
         if (bytesWritten != currOut.getCurrOffset() - startOffset) {
           throw new ISE("WTF? Perhaps there is some concurrent modification going on?");
         }
         if (bytesWritten != size) {
           throw new IOException(
               String.format("Expected [%,d] bytes, only saw [%,d], potential corruption?", size, bytesWritten)
-          );
+              );
```
cheddar: This whitespace appears to be off...

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -186,22 +196,108 @@ public void close() throws IOException
...
+        }
+
+        if(!writerCurrentlyInUse)
```
cheddar: Given that you just set `writerCurrentlyInUse` to `false` and that this object is not attempting to be thread-safe, I don't think this can ever be false.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -186,22 +196,108 @@ public void close() throws IOException
...
+        return  addToOffset(bytesWritten);
+      }
+
+      @Override
+      public int write(InputStream in) throws IOException
+      {
+        return addToOffset(Ints.checkedCast(java.nio.file.Files.copy(in, tmpFile.toPath())));
+      }
+
+      public int addToOffset(int numBytesWritten)
```
cheddar: This method has a really weird contract.  

1. Why is it `public`?
2. Why does it return its incoming argument without any adjustment?

The name of the method makes it seem like this would be returning the total offset instead of the size.  The methods in the non-delegate writer all do this call first with the name "verify" which makes it clear that the intent of the method is to do some state checks...

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -64,8 +66,11 @@
 
   private final List<File> outFiles = Lists.newArrayList();
   private final Map<String, Metadata> internalFiles = Maps.newTreeMap();
+  private List<File> files = Lists.newArrayList();
```
cheddar: How about renaming this to `completedFiles`.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
cheddar: Actually, how about a rename to `filesInProcess`?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -186,22 +196,108 @@ public void close() throws IOException
...
+      private int currOffset = 0;
+      private boolean open = true;
+      private final FileOutputStream out = new FileOutputStream(tmpFile);
+
+      @Override
+      public void close() throws IOException
+      {
+        open = false;
+        out.close();
+        files.add(tmpFile);
```
cheddar: Why wouldn't we remove the file from the `createFiles` or the `filesInProcess` here?

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -50,25 +52,68 @@ public void testSanity() throws Exception
...
-      for (int i = 0; i < 20; ++i) {
-        ByteBuffer buf = mapper.mapFile(String.format("%d", i));
-        Assert.assertEquals(0, buf.position());
-        Assert.assertEquals(4, buf.remaining());
-        Assert.assertEquals(4, buf.capacity());
-        Assert.assertEquals(i, buf.getInt());
-      }
-    }
+  @Test
+  public void testWhenFirstWriterClosedInTheLast() throws Exception
```
cheddar: Does "InTheLast" mean "AtTheEnd" or something else?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: 👍

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: Agree, It shouldn't be. Made the code changes to make it more meaningful.

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
akashdw: done

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -131,6 +136,11 @@ public void add(String name, List<ByteBuffer> bufferToAdd) throws IOException
 
   public SmooshedWriter addWithSmooshedWriter(final String name, final long size) throws IOException
   {
+    if(writerCurrentlyInUse)
+    {
+      return delegateSmooshedWriter(name , size);
```
leventov: space before ,

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -131,6 +136,11 @@ public void add(String name, List<ByteBuffer> bufferToAdd) throws IOException
 
   public SmooshedWriter addWithSmooshedWriter(final String name, final long size) throws IOException
   {
+    if(writerCurrentlyInUse)
```
leventov: Logic should go after argument check (if block below).

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -195,13 +206,95 @@ public void close() throws IOException
...
+    List<File> fileToProcess = new ArrayList<>(completedFiles);
+    completedFiles = Lists.newArrayList();
+    for(File file: fileToProcess)
+    {
+      add(file);
+      file.delete();
+    }
+  }
+
+  private SmooshedWriter delegateSmooshedWriter(final String name,final long size) throws IOException
```
leventov: space after ,

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -195,13 +206,95 @@ public void close() throws IOException
...
+
+      @Override
+      public void close() throws IOException
+      {
+        open = false;
+        out.close();
+        completedFiles.add(tmpFile);
+        filesInProcess.remove(tmpFile);
+
+        if(!writerCurrentlyInUse ) {
```
leventov: spacing

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -195,13 +206,95 @@ public void close() throws IOException
...
+      add(file);
+      file.delete();
+    }
+  }
+
+  private SmooshedWriter delegateSmooshedWriter(final String name,final long size) throws IOException
+  {	
+    final File tmpFile = new File(baseDir, name);
+    filesInProcess.add(tmpFile);
+    return new SmooshedWriter () {
```
leventov: no space before (), { on the new line

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: 👍

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: done

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: done

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: done

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
akashdw: fixed.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -195,13 +206,95 @@ public void close() throws IOException
...
+      }
+    };
+  }
+
+  private void mergeWithSmoosher() throws IOException
+  {
+    //get processed elements from the stack and write.
+    List<File> fileToProcess = new ArrayList<>(completedFiles);
+    completedFiles = Lists.newArrayList();
+    for(File file: fileToProcess)
```
akashdw: fixed

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -195,13 +206,95 @@ public void close() throws IOException
...
       }
     };
+
   }
 
   @Override
   public void close() throws IOException
   {
+    //book keeping checks on created file.
+    if(!completedFiles.isEmpty())
```
akashdw: fixed.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -195,13 +206,95 @@ public void close() throws IOException
...
+    {
+      for(File file: completedFiles)
```
akashdw: done.

akashdw:
@logarithm @leventov @cheddar Thanks for reviewing the PR. I addressed all the comments and updated the pr. My Eclipse auto-formatter was turned off which caused the formatting issues.

cheddar:
👍

leventov:
Please don't merge until the end of this week, I want to review this.

akashdw:
@leventov Thanks for looking into this PR, I'm using https://github.com/druid-io/druid/blob/master/eclipse_formatting.xml to format code. I'll update my local settings if that setting got updated.
I updated the code and addressed all the formatting comments, not sure why github is not showing the updated code in comment section, however files changes section is showing formatted code.

akashdw:
@leventov Updated the PR and added comments to explain code logic.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -63,8 +72,13 @@
 
   private final List<File> outFiles = Lists.newArrayList();
   private final Map<String, Metadata> internalFiles = Maps.newTreeMap();
+  // list of files completed writing content using delegated smooshedWriter.
```
leventov: I think it is better to write such comments as Javadoc `/** ... */`, then on automatic refactorings and reformattings IDEs will move comments along with fields. Javadoc comments are allowed for private fields.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -206,13 +231,132 @@ public void close() throws IOException
...
+  /**
+   * Merges temporary files created by delegated SmooshedWriters on to the main
+   * smoosh file.
+   *
+   * @throws IOException
+   */
+  private void mergeWithSmoosher() throws IOException
+  {
+    // Get processed elements from the stack and write.
+    List<File> fileToProcess = new ArrayList<>(completedFiles);
```
leventov: Maybe it can be just iteration over completedFiles (currently lines 251-256), then `completedFiles.clear()`.

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -206,13 +231,132 @@ public void close() throws IOException
...
 
   @Override
   public void close() throws IOException
   {
+    //book keeping checks on created file.
+    if (!completedFiles.isEmpty() || !filesInProcess.isEmpty())
+    {
+      for (File file: completedFiles)
+      {
+        file.delete();
```
leventov: Maybe merge completedFiles here using `mergeWithSmoosher()`, rather than silently delete?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
```diff
@@ -206,13 +231,132 @@ public void close() throws IOException
...
+      }
+      for (File file: filesInProcess)
```
leventov: Maybe throw some exception here, if filesToProcess is not empty? Because presence of files in `filesInProcess` means somebody opened a smoosherWriter but didn't close  it before closing FileSmoosher. It seems like a bug.

cheddar:
@leventov the idea to do it with composition is definitely interesting.  It poses a few problems as well because the only thing that currently has access to the `SmooshedWriter` to know if it is in use or not is the `FileSmoosher`.  So, the wrapping class would essentially also have to wrap the `SmooshedWriter` that is returned.

Also, there is a bit of state that the `FileSmoosher` necessarily owns, namely the directory that files are being created in.  Switching to composition would force this state to somehow be shared with the wrapping class (or the wrapping class would have to have yet another directory available to it for the storing of files).

I think that this code belongs inside `FileSmoosher` and going through the steps to make it work with composition will be just a bunch of code baggage without much benefit.

Can you elaborate on why you are afraid to change `FileSmoosher`?

src/main/java/com/metamx/common/io/smoosh/FileSmoosher.java
cheddar: There is an exception thrown just as soon as this for loop exits for exactly that reason. It's a bug if any files haven't yet been merged by the time the larger Smoosher gets closed (each of the individual Writers should have been closed and those should have caused everything to get merged).

leventov:
@cheddar 

> It poses a few problems as well because the only thing that currently has access to the SmooshedWriter to know if it is in use or not is the FileSmoosher. So, the wrapping class would essentially also have to wrap the SmooshedWriter that is returned.

From the description of https://github.com/druid-io/druid/issues/3513 I understood that @akashdw is going to pass FileSmoosher somewhere while SmoosherWriter is open. AFAIU it's not recursion, i. e. FileSmoosher is passed to some different place. A wrapper class could essentially have `delegateSmooshedWriter()` method public (however I would name it differently, somehow involving adjective "delayed"), and call this method from that other place. Don't need to track the FileSmoosher itself. I written this already in other words:

> I think this wrapper class could have specialized methods for writing to the delegate fileSmoosher directly and to temp files, because you know what do you need in the call sites in Druid. So the tricky writerCurrentlyInUse logic is not needed, I believe.

---

> Also, there is a bit of state that the FileSmoosher necessarily owns, namely the directory that files are being created in. Switching to composition would force this state to somehow be shared with the wrapping class (or the wrapping class would have to have yet another directory available to it for the storing of files).

The only bit of state that new logic accesses is baseDir, to create temp files. However temp files don't need (and actually shouldn't) be created in the smoosher directory itself. They should better be created via Java standard, or Guava's utility methods for creating temp files, which end up placing it in java.io.temp dir or /tmp, which is likely to be mounted to tmpfs, while the smoosher directory is less likely to be there (?)

---

So, I don't see the new logic is tightly coupled with the FileSmoosher state. Also a couple of weeks ago you mentioned that it's better to add code to druid than java-util. This also the case, I suppose.

leventov:
@cheddar Also, I think it's quite likely that  @akashdw's proposed design on adding functionality to Druid which he wants to add will be changes sooner or later. Maybe even as soon as on the PR review stage. So that this change is no longer needed, for whatever reason. But we will need to carry it because it's a library. While logic from Druid core could be removed pretty freely, because it is not a library. Also don't forget about boilerplate of having separate PRs and release cycle of java-util and Druid...

cheddar:
Ok, we will just move this to be part of the larger PR in druid.

cheddar:
Btw,

> They should better be created via Java standard, or Guava's utility methods for creating temp files, which end up placing it in java.io.temp dir or /tmp, which is likely to be mounted to tmpfs, while the smoosher directory is less likely to be there (?)

We cannot and should not do that.  By writing into random temporary locations, you are placing extra data in arbitrary places that the users do not have strong control over.  The directories used inside of Druid for storing these things have a very specific lifecycle and get cleaned up according to that lifecycle.  Reusing those directories is important and using tmp locations is a non-starter (and often leads to orphaned files and users not understanding why things have filled  up).

leventov:
@cheddar thanks, this makes sense. So, smooshed file directories are temporal as a whole? And could be located in tmpfs?

So if "MultiSmoosher" need to access FileSmoosher's baseDir, it could be passed along with FileSmoosher in constructor, or `getBaseDir()` accessor could be added to FileSmoosher.

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -50,26 +52,71 @@ public void testSanity() throws Exception
...
+      final SmooshedWriter writer = smoosher.addWithSmooshedWriter(String.format("%d", 19), 4);
+
+      for (int i = 0; i < 19; ++i) {
+        File tmpFile = File.createTempFile(String.format("smoosh-%s", i), ".bin");
+        Files.write(Ints.toByteArray(i), tmpFile);
+        smoosher.add(String.format("%d", i), tmpFile);
+        if (i==10)
+        {
+          writer.write(ByteBuffer.wrap(Ints.toByteArray(19)));
+          CloseQuietly.close(writer);
```
drcrallen: why quietly?

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -50,26 +52,71 @@ public void testSanity() throws Exception
...
+      final SmooshedWriter writer = smoosher.addWithSmooshedWriter(String.format("%d", 19), 4);
+      writer.write(ByteBuffer.wrap(Ints.toByteArray(19)));
+
+      for (int i = 0; i < 19; ++i) {
+        File tmpFile = File.createTempFile(String.format("smoosh-%s", i), ".bin");
+        Files.write(Ints.toByteArray(i), tmpFile);
+        smoosher.add(String.format("%d", i), tmpFile);
+        tmpFile.delete();
+      }
+      CloseQuietly.close(writer);
```
drcrallen: why quietly?

src/test/java/com/metamx/common/io/smoosh/SmooshedFileMapperTest.java
```diff
@@ -150,4 +197,26 @@ public void testDeterministicFileUnmapping() throws IOException
...
+      for (int i = 0; i < 20; ++i) {
+        ByteBuffer buf = mapper.mapFile(String.format("%d", i));
+        Assert.assertEquals(0, buf.position());
+        Assert.assertEquals(4, buf.remaining());
+        Assert.assertEquals(4, buf.capacity());
+        Assert.assertEquals(i, buf.getInt());
+      }
+    }
+  }
+}
```
drcrallen: Missing newline

drcrallen:
Thanks!

PR #56, opened 2016-10-04
Resolve most HuntBugs warnings (including two bugs fixed)

leventov:
- bug fix 1: https://github.com/metamx/java-util/pull/56/files#diff-aba6de6f4670f473afd1ae9ba6ef209bR86 (don't allocate `new Random()` each time we need a random value).
 - bug fix 2: https://github.com/metamx/java-util/pull/56/files#diff-be255ea78e884c6ed2e259dfd9528b94R43, Comparators.inverse() is broken if the source comparator returns `Integer.MIN_VALUE`.

### DIFF
--- a/src/main/java/com/metamx/common/RetryUtils.java
+++ b/src/main/java/com/metamx/common/RetryUtils.java
@@ -23,6 +23,7 @@ import com.metamx.common.logger.Logger;
 
 import java.util.Random;
 import java.util.concurrent.Callable;
+import java.util.concurrent.ThreadLocalRandom;
 
 public class RetryUtils
 {
@@ -82,7 +83,7 @@ public class RetryUtils
   {
     final long baseSleepMillis = 1000;
     final long maxSleepMillis = 60000;
-    final double fuzzyMultiplier = Math.min(Math.max(1 + 0.2 * new Random().nextGaussian(), 0), 2);
+    final double fuzzyMultiplier = Math.min(Math.max(1 + 0.2 * ThreadLocalRandom.current().nextGaussian(), 0), 2);
     final long sleepMillis = (long) (Math.min(maxSleepMillis, baseSleepMillis * Math.pow(2, nTry - 1))
                                      * fuzzyMultiplier);
     if (quiet) {
--- a/src/main/java/com/metamx/common/RetryUtils.java
+++ b/src/main/java/com/metamx/common/RetryUtils.java
@@ -21,7 +21,6 @@ import com.google.common.base.Predicate;
 import com.google.common.base.Throwables;
 import com.metamx.common.logger.Logger;
 
-import java.util.Random;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ThreadLocalRandom;
 
--- a/src/main/java/com/metamx/common/StreamUtils.java
+++ b/src/main/java/com/metamx/common/StreamUtils.java
@@ -132,9 +132,9 @@ public class StreamUtils
   public static long copyWithTimeout(InputStream is, OutputStream os, long timeout) throws IOException, TimeoutException
   {
     byte[] buffer = new byte[DEFAULT_BUFFER_SIZE];
-    int n = 0;
+    int n;
     long startTime = System.currentTimeMillis();
-    long size = 0l;
+    long size = 0;
     while (-1 != (n = is.read(buffer))) {
       if (System.currentTimeMillis() - startTime > timeout) {
         throw new TimeoutException(String.format("Copy time has exceeded %,d millis", timeout));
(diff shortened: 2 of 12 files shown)


logarithm:
Are these changes critical and not applying them may potentially lead to some errors? I would prefer not to change core code without evident reasons.

leventov:
@logarithm changes include two bug fixes: https://github.com/metamx/java-util/pull/56/files#diff-aba6de6f4670f473afd1ae9ba6ef209bR86 and https://github.com/metamx/java-util/pull/56/files#diff-be255ea78e884c6ed2e259dfd9528b94R43.

This PR is a preparation step before including huntbugs (findbugs) as regular steps into Maven build, that will help to prevent introducing some errors in the future.

logarithm:
Thanks! Descriptive comments to PR very useful for reviewers and posterity.

src/main/java/com/metamx/common/RetryUtils.java
```diff
@@ -23,6 +23,7 @@
 
 import java.util.Random;
```
logarithm: This import is unused now

src/main/java/com/metamx/common/guava/LimitedYieldingAccumulator.java
```diff
@@ -18,6 +18,7 @@
 
 /**
  */
+@Deprecated
```
logarithm: Why this class is marked as `Deprecated`?

src/main/java/com/metamx/common/StreamUtils.java
```diff
@@ -132,9 +132,9 @@ public static long copyAndClose(InputStream is, OutputStream os) throws IOExcept
   public static long copyWithTimeout(InputStream is, OutputStream os, long timeout) throws IOException, TimeoutException
   {
     byte[] buffer = new byte[DEFAULT_BUFFER_SIZE];
-    int n = 0;
+    int n;
     long startTime = System.currentTimeMillis();
-    long size = 0l;
+    long size = 0;
```
drcrallen: Out of curiosity, why is `0` better than `0L`?

src/main/java/com/metamx/common/guava/Comparators.java
```diff
@@ -40,7 +40,7 @@
       @Override
       public int compare(T t, T t1)
       {
-        return - baseComp.compare(t, t1);
+        return baseComp.compare(t1, t);
```
drcrallen: Can you add a unit test for this one?

src/main/java/com/metamx/common/guava/LimitedYieldingAccumulator.java
leventov: This class uses volatile internally, for counter, but it's not thread-safe, because volatile is used improperly. This class is not used neither in druid nor in metamx codebase. (There is a private inner class with the same name in LimitedSequence, which is used; apparently this top-level LimitedYieldingAccumulator is just forgotten to be removed during some refactoring.)

Since java-util has version 0.x I think it's ok to remove this class and update minor version (27 -> 28). What do you think?

src/main/java/com/metamx/common/StreamUtils.java
drcrallen: Ok sure. I think druid uses `0L` when possible but it is not strictly enforced

src/main/java/com/metamx/common/RetryUtils.java
leventov: This is the reason why I want to introduce checkstyle.. https://github.com/druid-io/druid/pull/3551

src/main/java/com/metamx/common/StreamUtils.java
leventov: Let's fix that when this rule will be enforced on Checkstyle level.

src/main/java/com/metamx/common/guava/LimitedYieldingAccumulator.java
leventov: Left the class, added a comment.

src/main/java/com/metamx/common/RetryUtils.java
leventov: Removed

src/main/java/com/metamx/common/parsers/JSONPathParser.java
```diff
@@ -60,7 +59,6 @@
    */
   public JSONPathParser(List<FieldSpec> fieldSpecs, boolean useFieldDiscovery, ObjectMapper mapper)
   {
-    this.fieldSpecs = fieldSpecs;
```
drcrallen: If this is not used, can we `@Deprecated` this constructor signature and make one that does not have `fieldSpecs`?

leventov: @drcrallen fieldsSpecs are used in the next line, to derive some other field, while are not needed as a field itself.

drcrallen: ah I see, missed that thanks.

PR #58, opened 2017-02-08
[BACKEND-644] Updated dependencies to latest compatible versions

logarithm:
for example guava 21.0 require Java 8 but we are still on Java 7

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,7 +86,7 @@
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>20.0</version>
+            <version>16.0.1</version>
         </dependency>
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -70,32 +70,38 @@
...
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
         <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
-            <version>16.0.1</version>
+            <version>20.0</version>
```
drcrallen: this breaks hadoop compat

logarithm: Updating of guava brakes hadoop because it depends on some removed from guava functions? Rolled back version for guava library.

logarithm:
@drcrallen Joda date time new version contains actual time zones information. Slf4J contains some important bug fixes and performance improvements. And so on.

Another goal of dependencies update is to keep versions up to date. For example scala 2.12 introduce performance improvements and heavily use Java 8 features, but hence depends on at least 8 version of Java. Now it too soon to move to the new scala version, but if we continue to keep dependencies outdated then migration to 2.12 in future might be much bigger problem. Of course java-util is not scala library, but a lot of scala code depends on this library.

PR #72, opened 2017-11-01
Do not log error messages if "exception" is less than minimum timeout.

xanec:


src/main/java/com/metamx/emitter/core/HttpPostEmitter.java
```diff
@@ -627,7 +627,11 @@ public boolean apply(Throwable e)
         return false;
       }
       catch (Exception e) {
-        log.error(e, "Failed to send events to url[%s]", config.getRecipientBaseUrl());
+        if (e == timeoutLessThanMinimumException) {
```
drcrallen: the normal way to do this is to have a specific Exception type that is handled independently, is it always guaranteed to have object equality?

xanec: I believe this is the specific `Exception` object used to throw when the timeout is less than the minimum and exit the retry:
https://github.com/metamx/java-util/blob/e935f9f9d8d12889aa92b99f482451528e176e3a/src/main/java/com/metamx/emitter/core/HttpPostEmitter.java#L643-L645

The comparison is also used above:
https://github.com/metamx/java-util/blob/e935f9f9d8d12889aa92b99f482451528e176e3a/src/main/java/com/metamx/emitter/core/HttpPostEmitter.java#L615-L617

I feel that this is technically not an error because we simply chose not to try sending it as the timeout is too low.

PR #74, opened 2017-12-21
reduce noise in the logs

himanshug:


leventov:
@himanshug please start the process of moving java-util into Druid.

himanshug:
@leventov java-util was already moved to https://github.com/druid-io/druid/tree/master/java-util .
http-client was in a separate project when that happened. I think you mean moving http-client also there.

Last we discussed, @drcrallen proposed that these moves should be done on a need basis. not sure if things have changed now and there is more agreement on that. @drcrallen are you ok with the moving http-client code to java-util folder in Druid considering it is no longer a standalone project now?

leventov:
This module separation inside Druid doesn't make any sense, it's a "dept". We should merge druid-api, druid-processing, common, java-util, extendedset, bytebuffer-collections and maybe some other modules into just one "druid-core".

himanshug:
moved to https://github.com/druid-io/druid/pull/5302

